Issue #3, opened 2024-04-29
Packaging the tests


jo-basevi:
Package up the tests so they can be installed via `conda`/`pip` and versioned. 

It is possible to pass a package to pytest `pytest --pyargs my-package` which runs test code inside that package: https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-as-part-of-application-code.
However, will need to test that custom command line args such as `--control-path` , `--output-path` etc still work (see https://stackoverflow.com/questions/41270604/using-command-line-parameters-with-pytest-pyargs). Otherwise will need to create wrapper commands that call pytest. (Will also need to restructure code into `src` and `test`, and add pyproject.toml etc.)

It will also need a think of how it is deployed in CI tests that is run on NCI. As the tests are currently run using the deployed `payu` module (https://github.com/ACCESS-NRI/payu-condaenv).  This package can be added to this conda environment but then the version in the payu conda environment is fixed - which could be fine when the package is in a stable state or the environment can be redeployed with an updated version.

jo-basevi:
Versioning tests were discussed in this issue https://github.com/ACCESS-NRI/access-om2-configs/issues/61, so older configurations can run earlier versions of test code.

jo-basevi:
Also add CI to test packaging, and also CI to run any local tests

jo-basevi:
Could use `pip install PACKAGE_NAME` as part of CI tests in `access-om2-configs` and `reproducibility`. A potential issue is in repro tests run on NCI, if there were two concurrent `repro` tests which were installing different versions

PR #19, opened 2024-05-06
Add AccessOm3 model class

### LINKED ISSUE #7: access-om3 tests
I [wrote a very simple `access-om3` model for these tests](https://github.com/dougiesquire/MOM6-CICE6/tree/main/test) when I was playing about with the repro CI workflows a month or so ago. Should I move this here, or is there still some dust that needs to settle?

dougiesquire:
This PR adds a very basic `AccessOm3` model class. I wrote this a while ago and I see that in the meantime a `check_checksums_over_restarts` method has been added to the `AccessOm2` class. From a very quick glance, it looks like the `AccessOm3` implementation of this method would be the same. So could this method implementation be moved to the `Model` base class?

I haven't added any tests to the `test` directory but probsibly should. Lemme know.

Closes #7


models/accessom3.py
```diff
@@ -0,0 +1,85 @@
...
+            output_filename = output_directory / 'ocean.stats'
+        else:
+            output_filename = self.output_file
+
+        # ocean.stats is used for regression testing in MOM6's own test suite
+        # See https://github.com/mom-ocean/MOM6/blob/2ab885eddfc47fc0c8c0bae46bc61531104428d5/.testing/Makefile#L495-L501
+        # Here we calculate the md5 hash of ocean.stats
+        with open(output_filename, 'rb') as f:
+            contents = f.read()
+            md5_hash = hashlib.md5(contents).hexdigest()
```
aidanheerdegen, line 68: Is there any chance checksumming the whole file will result in false positives/negatives, if say formatting changes?

aidanheerdegen: Also, is there any value in granularity in the tests? So knowing if particular tests failed? Or if they have changed, but only a little bit?

models/accessom3.py
```diff
@@ -0,0 +1,85 @@
+"""Specific Access-OM3 Model setup and post-processing"""
+
+import hashlib
+from pathlib import Path
+from payu.models.cesm_cmeps import Runconfig
+from typing import Dict, Any
+
+from models.model import Model
+
+BASE_SCHEMA_URL = "https://raw.githubusercontent.com/ACCESS-NRI/schema/main/au.org.access-nri/model/access-om2/experiment/reproducibility/checksums"
```
aidanheerdegen, line 10: I'm kinda surprised (in a good way) that the schema is so versatile it can be (ab)used in this way.

aidanheerdegen (approved):
I'm approving because why not? Something is better than nothing, and this is something.

aidanheerdegen:
This is the PR I mentioned in the stand up this morning @jo-basevi. I've tagged you for a review.

jo-basevi (dismissed):
Yeah, maybe the access-om2 schema could be the default checksum schema? And then it would make sense for `check_checksums_over_restarts` be the default method in the `Model` base class? Or alternatively, `check_checksums_over_restarts` be an external function in `accessom2` that can be imported to `accessom3`

If there's an example file for `ocean.stats` and an expected checksum value for it, I am happy to add in a test for it. I had a quick look at the `test` folder and I probably need to do a quick rewrite of it to generalise it anyway. I can either add to this PR or create a subsequent PR once this is merged?

models/accessom3.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if years == months == 0:
+            freq = "nseconds"
+            n = str(seconds)
+        elif seconds == 0:
+            freq = "nmonths"
+            n = str(12 * years + months)
+        else:
+            raise NotImplementedError(
+                f"Cannot specify runtime in seconds and year/months at the same time")
```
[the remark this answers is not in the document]
dougiesquire: I think there's a restriction in `access-om2` that only one of years, months or seconds is non-zero (e.g. see [here](https://github.com/COSIMA/1deg_jra55_ryf/blob/c7f15f671b3417c52e8ecd77d5fe9b4d38429452/accessom2.nml#L18))

models/accessom3.py
dougiesquire: > Is there any chance checksumming the whole file will result in false positives/negatives, if say formatting changes?

Yes, in this case the reference checksums would need to be regenerated I suppose. I'm not immediately sure what we could do instead. Extracting and checking contents would presumably also given bad behaviour if the formatting changes. Note, I chose to use the `ocean.stats` file here because that's what's used in MOM6's own regression testing, but there may be better outputs to check.

dougiesquire: > Also, is there any value in granularity in the tests? So knowing if particular tests failed? Or if they have changed, but only a little bit?

Maybe in the long run? I would also like to include output from other components, so this is definitely just a basic starting point.

Regarding small changes, I think any change means the model is not reproducing so should fail. But in the future we could include messaging about "how much" it's not reproducing.

models/accessom3.py
dougiesquire: Yeah, I chose the use the same schema mainly because it was easiest/fastest at the time. But I think the schema as it currently exists would work for most/all models: did any of the strings in this list change?

If we do decide to use this schema generally it may be worth renaming because users may want to put strings other than checksums in the list, e.g. values such as global sums (or can I still call that a "checksum"??)

dougiesquire:
> Yeah, maybe the access-om2 schema could be the default checksum schema? And then it would make sense for check_checksums_over_restarts be the default method in the Model base class? Or alternatively, check_checksums_over_restarts be an external function in accessom2 that can be imported to accessom3

Sounds good to me 😄

> If there's an example file for ocean.stats and an expected checksum value for it, I am happy to add in a test for it.

I've added an example `ocean.stats` and `access-om3-checksums-1-0-0.json` to `test/resources`. Would be great if you could write/refactor the tests thanks!

models/accessom3.py
aidanheerdegen: > If we do decide to use this schema generally it may be worth renaming because users may want to put strings other than checksums in the list, e.g. values such as global sums (or can I still call that a "checksum"??)

Yeah I had thought if it was more generally useful then it (c|s)hould have a less specific name.

models/accessom3.py
aidanheerdegen: > Yes, in this case the reference checksums would need to be regenerated I suppose. I'm not immediately sure what we could do instead. Extracting and checking contents would presumably also given bad behaviour if the formatting changes. Note, I chose to use the `ocean.stats` file here because that's what's used in MOM6's own regression testing, but there may be better outputs to check.

Having looked at the `ocean.stats` file I'd say it would be a little more robust to extract each of the fields and save them individually. Could use the header strings as keys.

This would have the advantage of being invariant/insensitive to *some* formatting changes. If you removed the strings from the values in the third line and read them in as floats it would also make the checksum part invariant to changes in whitespace, but potentially not to changes in the formatting of the values themselves.

In the repro CI we're using the state of the checksum reproducibility to determine when to change major version: when the model is no longer reproducing. It would be bad to get false positives and change the major version when the model was still reproducing previous answers.

dougiesquire: Done in [0313093](https://github.com/ACCESS-NRI/model-config-tests/pull/19/commits/0313093d459eb9a3b91e3f6fffa52b19602f47d3). In doing this it became clear that something like this is actually needed for the restart repro tests to work so thanks for the suggestion :)

test/resources/ocean.stats
```diff
@@ -0,0 +1,3 @@
+  Step,       Day,  Truncs,      Energy/Mass,      Maximum CFL,  Mean Sea Level,  Total Mass,  Mean Salin, Mean Temp, Frac Mass Err,   Salin Err,    Temp Err
+            [days]                 [m2 s-2]           [Nondim]       [m]             [kg]         [PSU]      [degC]       [Nondim]        [PSU]        [degC]
```
aidanheerdegen, line 2: The units line isn't comma separated! This is bad.

Out of your control, but still bad.

test/resources/access-om3-checksums-1-0-0.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema_version": "1-0-0",
+  "output": {
+    "ocean.stats": [
+      "d2f2f18e0f688c1717b52acf78c0793f"
```
aidanheerdegen, line 5: This needs to be updated to reflect the changed logic in `extract_checksums`.

aidanheerdegen (commented):
A couple more comments, didn't want to push approve until there were tests.

test/resources/access-om3-checksums-1-0-0.json
dougiesquire: Done in [1671601](https://github.com/ACCESS-NRI/model-config-tests/pull/19/commits/1671601668c2326961370f8e00cd6629669345bd)

jo-basevi:
Firstly, sorry for force pushing - I only realised after a `git push` that I needed to amend a commit message..

So I've updated the extract_checksums tests, it was very good because I noticed that schema has an error which will need to be fixed. I've added that and moving the schema to a more general location in this PR: https://github.com/ACCESS-NRI/schema/pull/24

Once, that is merged I can update the URL to point to a new location and commit, as the url is not used anywhere apart from the tests for extract_checksums under `test`.

As part of package tests issue #3, it'll will be good to add in CI to automatically run these tests so it is actually checking the schema validation.

models/accessom3.py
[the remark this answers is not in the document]
jo-basevi: Thanks, I've added in a check for that to the access-om2 class

dougiesquire:
> So I've updated the extract_checksums tests, it was very good because I noticed that schema has an error which will need to be fixed. I've added that and moving the schema to a more general location in this PR: https://github.com/ACCESS-NRI/schema/pull/24

Thanks @jo-basevi. Is the idea to merge https://github.com/ACCESS-NRI/schema/pull/24 first so that we can get the tests running here before merging? Or is the plan to update the schema in a subsequent PR (and merge this with failing tests)?

jo-basevi:
It would be great to merge the schema PR first. But if its needed to merge this to main asap, I can create a separate issue for updating schema url once the schema PR is merged.

dougiesquire:
> But if its needed to merge this to main asap

Nope, I'm not in any particular rush. I agree it would be good to have working tests before merging

dougiesquire (commented):
I can't approve my own PR, but I would if I could. Tests now run and LGTM - thanks @jo-basevi

jo-basevi (dismissed):
Just re-approving on the ACCESS-OM3 model changes

dougiesquire:
@aidanheerdegen looks like we need you to approve (if you're happy to)

models/accessom3.py
```diff
@@ -0,0 +1,90 @@
+"""Specific Access-OM3 Model setup and post-processing"""
+
+from collections import defaultdict
+import re
+from pathlib import Path
+from payu.models.cesm_cmeps import Runconfig
```
jo-basevi, line 6: Sorry, it only just occurred to me that importing payu might run into import errors in `access-om2-configs` QA tests CI https://github.com/ACCESS-NRI/access-om2-configs/blob/a57e7fd10c351edef7427ddf42c71ce53c24df27/.github/workflows/pr-1-ci.yml#L114-L143, as `payu` is not included `requirements.txt`. It would work ok on NCI for the repro tests, as they use payu conda environment in vk83. @aidanheerdegen we talked about earlier today about using separate python virtual environments on NCI to be able to install different versions of `model-config-tests` for potentially concurrent repro tests. Should a version of payu be included in that so the environment is separate from the payu conda environments and payu can be included in requirements.txt?

dougiesquire: Ah I (mis)read [this](https://github.com/ACCESS-NRI/model-config-tests/blob/b086103bb4a2ff87f7d6bb82cd767ab173b1e089/models/accessom2.py#L3-L4) to mean that Payu would always be available.

jo-basevi:
I'll rebase this branch onto changes from main as I imagine there's going to be a few merge conflicts

models/accessom3.py
jo-basevi: Yeah sorry about that, I've included payu as a dependency so it'll now be available

jo-basevi:
I ran a test on gadi to double check I didn't mess up the merge, and can confirm no errors when testing a released access-om2-configs configuration.

src/model_config_tests/models/accessom3.py
```diff
@@ -0,0 +1,90 @@
+"""Specific Access-OM3 Model setup and post-processing"""
+
+import re
+from collections import defaultdict
+from pathlib import Path
+from typing import Any
+
+from payu.models.cesm_cmeps import Runconfig
```
aidanheerdegen, line 8: Sorry to bring this up so late @jo-basevi, but maybe we don't want to do it this way?

If we need to reuse `Runconfig` then should it be in its own package?

Does something like `om3-utils` also need to reuse this or similar manipulation routines?

dougiesquire: `om3-utils` doesn't reuse this, but it does have similar functionality. We could use `om3-utils` instead, but is that any better?

jo-basevi: Maybe in future when `om3-utils` is packaged, could use reuse `om3-utils` instead. 

When running repro tests on gadi, if payu is a dependency it needs to be installed inside the test environment. Otherwise if `payu` is not a dependency, `payu` can be obtained from `g/data/vk83/apps/payu` modules, as it just needs the `payu run` command.

So I think it's fine to keep payu as a dependency for now. A benefit is that it keeps the test environment more isolated. But if the number of files of the test environments becomes a problem, then removing payu as a dependency could maybe help.

aidanheerdegen (approved):
LGTM

PR #22, opened 2024-05-17
Package model-config-tests

### LINKED ISSUE #3: Packaging the tests

jo-basevi:
- Restructured the code into `src` and `test`
- Add `model-config-tests` cmd to run pytests under `model_config_tests`
- Added CD: On push to tags build distribution and publish to PyPI (version would need to be updated in `pyproject.toml`, unless it is made dynamic like `payu` with versioneer?)
- Added CI: Runs local tests, and test building distribution

Note: Test code gets rewritten in #19, so changes under `tests/` will get rewritten.

Before merging: Need PRs in access-om2-configs and repro repositories pinning `model-config-tests` to a previous commit

Closes #3


jo-basevi:
I've used a bunch more of the example CI/CD actions here (https://github.com/ACCESS-NRI/dev-docs/wiki/Continuous-integration-and-deployment):
- adding pre-commit for linting / formatting, 
- uploading code coverage to codecov,
- Running tests with python versions 3.9-3.11

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+  pull_request:
+    branches:
+      - 'main'
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
CodeGat, line 13: We might want to specify a particular version of python - we seem to be using `3.11` a lot: 
```suggestion
      - uses: actions/setup-python@v5
        with:
          python-version: '3.11'
          cache: pip
```

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+    if: github.repository == 'access-nri/model-config-tests'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
```
CodeGat, line 27: This needs to be in quotes because otherwise it interprets it as a floating point rather than the required `string`:
```suggestion
          python-version: '3.11'
```

.github/workflows/CI.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Install build dependencies
+        run: python3 -m pip install --upgrade build
+
+      - name: Build distributions
+        run: python3 -m build
+
+      - name: Upload distribution artifact
+        uses: actions/upload-artifact@v4
+        with:
+          path: dist
```
CodeGat, line 39: We upload this artifact as part of CI, but we don't download it again during this run. Just making sure that we don't want to use it for anything in the later `tests` job. It's fine to just have it for provenance, but remember that workflow artifacts have a 90 day expiry.

src/model_config_tests/__main__.py
```diff
@@ -0,0 +1,18 @@
...
+import sys
+
+# Running pytests using --pyargs does not run pytest_addoption in conftest.py
+# Using workaround as described here:
+# https://stackoverflow.com/questions/41270604/using-command-line-parameters-with-pytest-pyargs
+HERE = os.path.dirname(__file__)
+
+
+def main():
+    import pytest
```
CodeGat, line 11: image
There are a whole bunch of these warnings [on the Files view](https://github.com/ACCESS-NRI/model-config-tests/pull/22/files) of this PR...what do they mean? Can it be turned off? :grin: Unless it's just because it's the initial run of the CodeCov stuff

CodeGat (commented):
Did my review - I assume the pre-commit hooks for `black` ran over this PR as well as there are a whole lot of `'foo'` to `"foo"` changes...in future the `black` reformatting might have to be a separate PR that runs through the whole project - that way we know that nothing will change except the formatting.

.github/workflows/CD.yml
```diff
@@ -0,0 +1,49 @@
+name: CD
+
+on:
+  push:
+    tags:
+      - '*'
```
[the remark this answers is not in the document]
jo-basevi: If I remove `- "*"`, my IDE gives an error that tags has to be an array

.github/workflows/CI.yml
jo-basevi: I think it's mostly there to test the `pypi-build` job in CD will work fine, so I could remove the upload part as probably it is unnecessary to store all these artefacts for 90 days.

src/model_config_tests/__main__.py
jo-basevi: Looks like codecov goes over every line changed in the PR, and displays if it has been covered by tests.. I mean this project has a very low test coverage so far, so it's probably good to have codecov reminder to have more tests in future. They can be disabled in the codecov config, or if in the files view, you can press `a` to toggle the annotations on/off (https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations)

.github/workflows/CD.yml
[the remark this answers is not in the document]
CodeGat: Interesting! Should be fine to keep it, then.

jo-basevi:
Thanks for the review @CodeGat! Yeah, that've would've been been good to have done it in a separate PR in future.. I think my goal in future is to set up `pre-commit` from start

CodeGat (dismissed):
Thanks, LGTM

jo-basevi:
I noticed come errors with edits I made to the pyproject.toml file. I thought it would be tidy to have local test settings configured in pyproject.toml but of course they also had an impact on pytests inside the package (that are run by `model-config-tests` cmd). So I've added a step to CI to test running `model-config-tests` command.  I've added that to a test that also `pip installs` `model_config_tests` from the distribution built for PyPI, so to also check that `model-config-tests` cmd has no runtime errors with missing dependencies..

CodeGat (dismissed):
Went through the additions and they look good. Keen to see how the matrixed code coverage stuff goes, I haven't used that before.

README.md
```diff
@@ -30,7 +37,7 @@ module load payu/1.1
```
aidanheerdegen, line 37: Should either update the `payu` version or just use the default (which should be latest)

src/model_config_tests/conftest.py
```diff
@@ -14,19 +14,19 @@ def output_path(request):
...
-    path = request.config.getoption('--output-path')
+    path = request.config.getoption("--output-path")
     if path is None:
         # Set default to /scratch/PROJECT/USER/test-model-repro/
-        project = os.environ.get('PROJECT')
-        user = os.environ.get('USER')
-        path = f'/scratch/{project}/{user}/test-model-repro'
+        project = os.environ.get("PROJECT")
+        user = os.environ.get("USER")
+        path = f"/scratch/{project}/{user}/test-model-repro"
```
aidanheerdegen, line 22: Maybe we should just use `$TMPDIR` environment variable? It resolves to `/scratch/$PROJECT/$USER/tmp` which is fine, and it would make it less NCI-centric for the future.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Setup Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+
+      - name: Install model_config_tests
+        run: python3 -m pip install ./dist/model_config_tests-*.tar.gz
+
+      - name: Run model-config-tests command
+        run: model-config-tests --collect-only -m "checksum"
```
aidanheerdegen, line 62: If we're running this on GitHub is`checksum` the tests we want?

src/model_config_tests/conftest.py
aidanheerdegen: That would require changing the docs too I guess.

aidanheerdegen (approved):
LGTM but questions ...

src/model_config_tests/conftest.py
jo-basevi: Oh awesome, I did not know that was a thing, I'll add that in

.github/workflows/CI.yml
jo-basevi: I am just testing for dependency runtime errors hence the --collect-only but it doesn't run the tests. Though I could just change it to `model-config-tests --collect-only` and it prints all the tests in the src code.

README.md
jo-basevi: I've updated the payu version to 1.1.3. I think I am going to rewriting this readme anyway soon depending on the outcome of this issue: https://github.com/ACCESS-NRI/reproducibility/issues/27

aidanheerdegen (approved):
LGTM

Issue #23, opened 2024-06-06
Move CI from `access-om2-configs`/`reproducibility` into this repository


CodeGat:
## Background
In order to add `ACCESS-ESM1.5` config CI (related issue here: https://github.com/ACCESS-NRI/access-esm1.5-configs/issues/15), we need to generify and consolidate the existing `access-om2-configs` workflows ([found here](https://github.com/ACCESS-NRI/access-om2-configs/tree/c30983ff077d522f39e0e4e13ff09f5dcd33dbe8/.github/workflows)) as well as the `reproducibility` workflows ([found here]( https://github.com/ACCESS-NRI/reproducibility/tree/5c3ca6f1fa9c85e5fe6f34e77bcc844a16e13c10/.github/workflows)). 

## Tasks
 - [x] `access-om2-configs`: Move `ci.schema.json` to `ACCESS-NRI/schema` and use the associated action in `ACCESS-NRI/schema` to validate each model-configs `ci.json`. See PRs https://github.com/ACCESS-NRI/access-om2-configs/pull/116 and https://github.com/ACCESS-NRI/schema/pull/32
 - [x] `access-om2-configs`: Move `pr-*.yml`, `schedule-*.yml` workflows to this repository. See https://github.com/ACCESS-NRI/access-om2-configs/pull/117 and https://github.com/ACCESS-NRI/model-config-tests/pull/24
 - [x] `access-om2-configs`: Open a PR that makes all `call-pr-*.yml` workflows reference the above moved workflows. Also add calls to `pr-2-confirm.yml` and `schedule-*.yml` workflows. **Important**: This change will have to be reflected in *all* config branches in `access-om2-configs` - `release-*` and `dev-*`, as well as `main`! See https://github.com/ACCESS-NRI/access-om2-configs/pull/117 for changes in `main`, changes will need to be force-propagated to every branch by an admin
 - [x] `reproducibility`: Move `checks.yml` and `generate-initial-checksums.yml` to this repository, consider archiving/moving this repo (ask @aidanheerdegen, [issue here](https://github.com/ACCESS-NRI/reproducibility/issues/30) ). See https://github.com/ACCESS-NRI/reproducibility/pull/29 and https://github.com/ACCESS-NRI/model-config-tests/pull/26
 - [x] `model-config-tests`: Make sure that there is no model-config-specific logic in the workflows in this repository
 - [x] `model-config-tests`: Check GitHub Environment variables and links
[...]

PR #26, opened 2024-06-11
Addition of generic workflows from `ACCESS-NRI/reproducibility`

CodeGat:
In this PR:
* Add config-generate-checksums.yml, config-pr-checks.yml from ACCESS-NRI/reproducibility#29

References #23

### DIFF
--- a/.github/workflows/config-generate-checksums.yml
+++ b/.github/workflows/config-generate-checksums.yml
@@ -1,0 +1,193 @@
+name: Initial Checksums
+on:
+  workflow_call:
+    inputs:
+      model-name:
+        type: string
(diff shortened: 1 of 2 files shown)


jo-basevi (approved):
Looks good to me. Is the documentation on these workflows (that was in reproducibility README) being added in subsequent PR's?

CodeGat:
I'll be taking inspiration from them, yes. I'm not certain of their final form as maybe certain parts of the reusable workflows can be more tightly integrated in the other workflows. See https://github.com/ACCESS-NRI/model-config-tests/pull/24 for the final form of these things

PR #27, opened 2024-06-19
Scheduled checks jq tags fix

jo-basevi:
Fix jq command in `.github/workflows/config-schedule-1-ci.yml` so output of tags to run scheduled checks on is an array

Failing scheduled check run in access-om2-configs repository: https://github.com/ACCESS-NRI/access-om2-configs/actions/runs/9576132337/job/26402116349


CodeGat (approved):
Looks good!!

PR #28, opened 2024-07-02
Add access class for access-esm repro testing

jo-basevi:
- Added an `access` model class (`access` is the name of the payu driver for the `access-esm1.5` model).
- Refactor extracting mom5 checksums to a separate file and function so it can be reused in `access-om2` and `access`. I used just a helper function, rather than model class of mom5 with `extract_checksums()` method to easily allow for different checksums be added to access-om2/access models in future.
- Added a test `access.out` and expected checksums to test resources so it passes the existing `extract_checksums()` tests for each model class.
- Set default model runtime for access to 24hr for `access` (as the minimum runtime is day), and keep access-om2/om3 to 3hrs. This default runtime is used in the historical repro checksums tests.

- Refactored the `test_bit_repro_historical` a little as the test was getting long with the added logic..

- Added a start of an integration-like test for `test_bit_repro_historical`. This is still draft while I am finishing off adding tests. But if there's any feedback on the changes is welcome


jo-basevi:
- Added more tests for `test_bit_repro_historical`. It's somewhat ironic having a lot of test code for tests.. The tests probably can be improved and broken up to test more smaller functions that are moved outside of `test_bit_reproducibility.py` and just have one or 2 integration style tests. In future, I want to test the other repro and config pytests, and other models rather than just `access`. Though that might be moving out of the scope of this PR.
- Added a quick improvement to the logging when the payu run fails in `test_bit_repro_historical`. I could add more logging, and use the `logging` library but I'll save that for an PR of this issue: https://github.com/ACCESS-NRI/model-config-tests/issues/5

src/model_config_tests/models/access.py
```diff
@@ -0,0 +1,77 @@
...
+import yaml
+
+from model_config_tests.models.model import SCHEMA_VERSION_1_0_0, Model
+from model_config_tests.models.mom import mom5_extract_checksums
+
+# Default model runtime (24 hrs)
+DEFAULT_RUNTIME_SECONDS = 86400
+
+
+class Access(Model):
```
CodeGat, line 15: Just for my own understanding, why is `Access` `ACCESS-ESM1.5` specific? It would seem to imply a suite of ACCESS models rather than just ESM1.5.

CodeGat (approved):
Wow, from all of that I only have one comment :smile: Looks great. 

Have you tested the test code?

src/model_config_tests/models/access.py
jo-basevi: I used `Access` as it is same name of the payu model driver used to run ACCESS-ESM1.5 (so that's the `model: access` in config.yaml). I could name the class differently to make it more clear, so something like `AccessEsm1p5` or `AccessEsm`? @aidanheerdegen would `access` be used for any other models? The payu code just refers to it as `The payu interface for the ACCESS coupled climate model`

jo-basevi:
I've tested the code with the pre-industrial configuration: https://github.com/ACCESS-NRI/access-esm1.5-configs/tree/pre-industrial, but using the executable paths from coecms: https://github.com/coecms/access-esm/blob/main/config.yaml, as the exe paths on the current  pre-industrial branch weren't working.

Though I think the pre-industrial configuration is getting updated with spack executables, so I'll be able to do another test run.

I've tested it with the broken 025 access-om2 configs tests to see what logs are printed when things go wrong..

jo-basevi:
Ok, so the spack executables in this PR: https://github.com/ACCESS-NRI/access-esm1.5-configs/pull/26, pass the historical bit reproducibility test using checksums from the `coecms` executables as a reference.

Restart reproducibility fails, because the model is not in a good state to restart from after a model runtime of one day - it's minimum runtime might be a month (?). Talked with @aidanheerdegen about whether restart reproducibility should be tested as part of Reproducibility CI checks for configs (as they fail for the access-om2-bgc configs too), and decided that bit historical reproducibility was sufficient testing for now. So I've changed the test marker to `checksum_slow`, and will create a separate issue for refactoring the restart repro tests. Does that sound an ok plan?

src/model_config_tests/models/access.py
aidanheerdegen: The driver could be used for other ACCESS models, but currently isn't as far as I know.

These tests are more specific to ESM1.5 I think, in that models can share a driver but have quite different configuration requirements.

So I'd be in favour of making it ESM1.5 specific.

aidanheerdegen:
> Ok, so the spack executables in this PR: [ACCESS-NRI/access-esm1.5-configs#26](https://github.com/ACCESS-NRI/access-esm1.5-configs/pull/26), pass the historical bit reproducibility test using checksums from the `coecms` executables as a reference.

Hey that's cool!

> Restart reproducibility fails, because the model is not in a good state to restart from after a model runtime of one day - it's minimum runtime might be a month (?). Talked with @aidanheerdegen about whether restart reproducibility should be tested as part of Reproducibility CI checks for configs (as they fail for the access-om2-bgc configs too), and decided that bit historical reproducibility was sufficient testing for now. So I've changed the test marker to checksum_slow, and will create a separate issue for refactoring the restart repro tests. Does that sound an ok plan?

I believe ESM1.5 has never been restart reproducible. @MartinDix mentioned that CABLE (the land model embedded in the UM in ESM1.5) also prevented restart reproducibility, so best to not have it turned on as a check. Which also means we can just use 1 day runs for historical reproducibility.

tests/test_test_bit_reprodubility.py
```diff
@@ -0,0 +1,293 @@
...
+    test_checksum = helper.output_path / "checksum" / "historical-24hr-checksum.json"
+    assert test_checksum.exists()
+
+    # Test that they are equal to checksums in resource folder
+    checksum_path = helper.resources_path / "checksums" / "1-0-0.json"
+    assert test_checksum.read_text() == checksum_path.read_text()
+
+
+def test_test_bit_repro_historical_access_no_model_output(tmp_dir):
+    """Check when a test exits, there are no checksums are in output directory
```
aidanheerdegen, line 268:
```suggestion
    """Check when a test exits, that there are no checksums in the output directory
```
Just a typo. Is that what you meant?

aidanheerdegen (dismissed):
Just a warning, this was a pretty light review, I didn't have time to check all the refactoring, but it looks great to me. Thanks for all the hard work!

src/model_config_tests/models/access.py
jo-basevi: I've renamed the model class to `AccessEsm1p5` so it's more clear it's configured for ESM1.5

tests/resources/access/output000/access.out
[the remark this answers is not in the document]
jo-basevi: I've removed a bunch of warnings from it so it's now `296K`

tests/test_test_bit_reprodubility.py
jo-basevi: Yes thanks, I've added that in

src/model_config_tests/test_bit_reproducibility.py
```diff
@@ -5,70 +5,141 @@
...
+        assert exp.model.output_exists(), (
+            "Output file for the model does not exist. "
+            "See the logs for more information on the experiment run"
+        )
 
-        # Check checksum against historical checksum file
-        hist_checksums = None
-        hist_checksums_schema_version = None
+        # Set the checksum output filename using the model default runtime
+        runtime_hours = exp.model.default_runtime_seconds // 3600
```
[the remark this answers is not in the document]
jo-basevi: Yeah thats a good idea. I've added those constants to the general `utils.py` file.

jo-basevi:
I reran the historical repro test with ACCESS-ESM1.5 pre-industrial config and an ACCESS-OM2 config, just to make sure the latest changes didn't break the tests.

src/model_config_tests/util.py
```diff
@@ -4,14 +4,19 @@
 import subprocess as sp
 import time
 
+# Time related constants
+SECONDS_IN_MINUTE = 60
+SECONDS_IN_HOUR = SECONDS_IN_MINUTE * 60
+SECONDS_IN_DAY = SECONDS_IN_HOUR * 24
```
CodeGat, line 10: I kinda feel like these would be `MINUTE_IN_SECONDS`, `HOUR_IN_SECONDS`, `DAY_IN_SECONDS`...Just so it's `<measure>_IN_<unit>` rather than `<unit>_IN_<measure>` - front-loading the important part, I think.

tests/resources/access/output000/access.out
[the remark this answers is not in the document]
CodeGat: :pray:

CodeGat (commented):
Just one thing... :smile:

src/model_config_tests/util.py
jo-basevi: That's a good point. I'll change those now

CodeGat (approved):

PR #30, opened 2024-07-09
Update QA executable paths tests

### LINKED ISSUE #18: Check executable paths match those deployed for specified model version
Once we have the ability to utilise `module load model/version` in `payu` configurations

https://github.com/ACCESS-NRI/model-config-tests/issues/10

we should have a QA configuration check that compares the executable paths in `manifests/exe.yaml` and check they match an install path in the `spack.locations.json` release artefact for the relevant model version

https://github.com/ACCESS-NRI/build-cd/issues/51

The release is available at a predictable URL for a given model version, e.g. 

https://github.com/ACCESS-NRI/ACCESS-OM2/releases/tag/2024.03.0

and the relevant artefact is just a path added to that url, e.g.

https://github.com/ACCESS-NRI/ACCESS-OM2/releases/download/2024.03.0/spack.location

(note that `spack.location.json` was not generated for the above release, but will be available in the future)

### LINKED ISSUE #29: Remove general config test for exe paths
In the eventual next release of payu, model executables can be set using modules. The current QA config tests check for absolute paths - and also that the exe paths are in the manifest: 

https://github.com/ACCESS-NRI/model-config-tests/blob/154705ee830c9a046dee5eb6903fb26126fc91c1/src/model_config_tests/test_config.py#L78-L109

These tests will either need to be removed or updated

jo-basevi:
In an upcoming version of payu, model modules with be loaded using `modules use/load` in payu config.yaml files, and exe paths for the model/submodel will then be determined by payu using the PATHs added by the module. 

- Removed QA tests for absolute executables paths in config.yaml files, and checking full paths in manifests
- Instead find the module version in config.yaml, and use that to find the released `spack.location` file. Then check the executable paths in manifests match an install path in the `spack.location`, as described in #18
- As ACCESS-OM2 has different repository and module for BGC configurations, and the only way to determine that has been checking if "bgc" is in the branch name, the above test has been moved into model specific tests.
- Added a start of ESM1.5 specific QA test class - under `access_esm1p5` marker
- Added tests for access-om2 configurations - added skeleton files on two configurations to test low/high-res + bgc/non-bgc specific tests. While I was at it, also added a test for bit repro historical for access-om2 using a test resource configuration to check runtime and the checksum file were set as expected.
- Added test for module in released configurations be `/g/data/vk83/modules/access-models`( For example for access-om2 rather than `/g/data/vk83/apps/spack/0.20/release/modules/linux-rocky8-x86_64`)

This requires a new release of payu - as need changes to find exe paths using modules.

This test will fail if using prerelease modules, as there is not a released `spack.location` file. 

Closes #18, Closes #29

### DIFF
--- a/src/model_config_tests/test_config.py
+++ b/src/model_config_tests/test_config.py
@@ -22,7 +22,7 @@ SCHEMA_COMMIT = "4b7207e47afe402a732c58741ff66acc5f93b8cf"
 LICENSE = "CC-BY-4.0"
 LICENSE_URL = "https://creativecommons.org/licenses/by/4.0/legalcode.txt"
 
-# Releaase modules location on NCI
+# Release modules location on NCI
 RELEASE_MODULE_LOCATION = "/g/data/vk83/modules/access-models"
 
 
@@ -211,7 +211,9 @@ def check_manifest_exes_in_spack_location(
     model_module_name, model_repo_name, control_path, config
 ):
     """This compares executable paths in the executable manifest, and checks
-    they match an install path in the spack.location release artefact.
+    they match an install path in the spack.location release artefact. The
+    version defined in the module configuration in config.yaml, is used
+    to find the relevant release version.
 
     This is called in model-specific config tests.
 
@@ -227,28 +229,24 @@ def check_manifest_exes_in_spack_location(
         The contents of the config.yaml file
     """
     help_msg = (
-        "Expected module for the model is added to loaded modules in config.yaml. E.g.\n"
+        "Expected module for the model is added to loaded modules in config.yaml. "
+        "The module also requires a released version. E.g.\n"
         "   modules:\n"
         "     use:\n"
         f"       - {RELEASE_MODULE_LOCATION}\n"
         "     load:\n"
-        f"       - {model_module_name}/<version>"
+        f"       - {model_module_name}/<version>\n"
         "Model executable paths can then be filenames that found in paths added by loaded module"
     )
 
     # Check module is defined in configuration file
     assert "modules" in config and "load" in config["modules"], help_msg
     loaded_modules = config["modules"]["load"]
-    modules = [m for m in loaded_modules if m.startswith(model_module_name)]
+    modules = [m for m in loaded_modules if m.startswith(f"{model_module_name}/")]
     assert len(modules) == 1, help_msg
 
     # Extract out the version
-    module = modules[0]
-    assert "/" in module, (
-        f"Expected {model_module_name} module version to be specified in config.yaml, "
-        f"e.g. {model_module_name}/<version>"
-    )
-    _, module_version = module.split("/")
+    _, module_version = modules[0].split("/")
 
     # Use the module version to download spack.location file
     url = (
@@ -270,13 +268,14 @@ def check_manifest_exes_in_spack_location(
     for exe_path in exe_paths:
         install_path, exe_name = exe_path.split("/bin/")
         assert install_path in str(spack_location), (
-            "Expected exe path in exe manifest to match an install path in released spack_location"
+            "Expected exe path in exe manifest to match an install path in released spack.location "
             f"for {model_module_name}/{module_version}.\n"
-            f"Executable path: {exe_path}"
+            f"Executable path: {exe_path}\n"
             f"----spack.location---\n{spack_location}"
         )
 
         assert exe_name in config_exes, (
-            "Expected model/submodel exe value in config.yaml to be the name of the executable,"
-            f"as the full path is determined using PATHs added by {model_module_name} module"
+            f"Expected 'exe: {exe_name}' for model/submodel in config.yaml. "
+            "Only the name of the executable is needed, as the full path is "
+            f"determined by payu (which searches PATHs added by {model_module_name} module)"
         )
--- a/src/model_config_tests/test_config.py
+++ b/src/model_config_tests/test_config.py
@@ -187,7 +187,7 @@ class TestConfig:
         )
 
 
-def exe_manifest_fullpaths(control_path: Path):
+def read_exe_manifest_fullpaths(control_path: Path):
     """Return the full paths to the executables in the executable manifest file"""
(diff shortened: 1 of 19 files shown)


CodeGat:
Just with regards to the following:
> This test will fail if using prerelease modules, as there is not a released spack.location file.

There is also `spack.location` files for Prereleases as well :)

src/model_config_tests/test_config.py
```diff
@@ -223,3 +185,97 @@ def test_license_file(self, control_path):
...
+    assert len(modules) == 1, help_msg
+
+    # Extract out the version
+    _, module_version = modules[0].split("/")
+
+    # Use the module version to download spack.location file
+    url = (
+        f"https://github.com/ACCESS-NRI/{model_repo_name}/"
+        f"releases/download/{module_version}/spack.location"
+    )
```
CodeGat, line 255: Ah, now I understand why we don't do it for Prereleases - you're getting it from the repo not from the environment. If you did want to get it from the the environment, you would find the `spack.location` file in `/g/data/vk83(/prerelease)?/apps/spack/<version>/spack/var/spack/environments/<model>_<version>/spack.location`.

tests/test_test_access_om2_config.py
```diff
@@ -0,0 +1,102 @@
...
+        # Use target branch as can't mock get_git_branch function in utils
+        f"--target-branch release-1deg_jra55_ryf"
+    )
+
+    result = subprocess.run(shlex.split(test_cmd), capture_output=True, text=True)
+
+    # Expect test to have failed
+    assert result.returncode == 1
+    error_msg = "Expected exe path in exe manifest to match an install path in released spack.location"
+    assert error_msg in result.stdout
```
CodeGat, line 75: Good idea to assert the particular type of error - I wouldn't have thought of that :laughing: 

Just make sure that if you modify the error message in future PRs, to update it in the test too :)

CodeGat (commented):
Looks good. Just a few comments.

src/model_config_tests/test_config.py
jo-basevi: Ok, just thinking on how to get it from environment, we would need to know the spack version (e.g. `0.20`, `0.22`). The models used in configs might use different spack versions, so this might not be able to be a fixed Github variable.

If module use paths, were the the longer paths in `config.yaml, we could infer more information e.g.
```
modules:
 use:
    - /g/data/vk83/apps/spack/0.20/release/modules/linux-rocky8-x86_64
 load:
    - access-om2/2024.03.0
```
The test code could be able refer the spack versions from `0.20`, and that it's not `prerelease` with `/g/data/vk83/apps` rather than `/g/data/vk83/prerelease/apps`. The `<model>_<version>` from `access-om2/2024.03.0` will need to converted to `access-om2-2024_03_0` to build path: 
```
/g/data/vk83/apps/spack/0.20/spack/var/spack/environments/access-om2-2024_03_0/spack.location
```
But then this can the con of using `/g/data/vk83/apps/spack/0.20/release/modules/linux-rocky8-x86_64` over `/g/data/vk83/modules/access-models`. But I don't have any other ideas on how to otherwise find the spack version.. 

I know there's been some talk about opening PR in `<model>-configs` when there's a new pre-release model built so repro tests can be run. I am wondering for these PRs with pre-release models that only `repro` tests  are run, and that `config`/QA tests are for PRs that will be merged into `release-*` branches.  For example, say if there was a workflow that would  open a PR and dynamically configure `config.yaml` to use the prerelease module path and module version. It would/should fail the `config` tests, because the exe manifest has not been updated (it needs a `payu setup` on gadi to be run to update the manifest files). It also should fail some tests, because we want to ensure there's a released `spack.location`, e.g. we don't want to have `prerelease` module in released configurations.

Yeah I am not really too sure, what do you think?

jo-basevi: Oh wait... QA tests aren't run on gadi, so I wouldn't be able read `/g/data/vk83/apps/spack/0.20/spack/var/spack/environments/access-om2-2024_03_0/spack.location` anyway

src/model_config_tests/test_config.py
```diff
@@ -21,14 +22,8 @@
...
-
-@pytest.fixture(scope="class")
-def exe_manifest_fullpaths(control_path: Path):
-    manifest_path = control_path / "manifests" / "exe.yaml"
-    with open(manifest_path) as f:
-        _, data = yaml.safe_load_all(f)
-    exe_fullpaths = {item["fullpath"] for item in data.values()}
-    return exe_fullpaths
+# Release modules location on NCI
+RELEASE_MODULE_LOCATION = "/g/data/vk83/modules/access-models"
```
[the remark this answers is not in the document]
jo-basevi: Yeah, it is possible to add it in as an argument - see this commit on a different branch: https://github.com/ACCESS-NRI/model-config-tests/commit/12d772248e2b797a9e82e8fe0aa719c4f0367765
However QA config tests are not run using an environment yet. So, could add `Gadi` as the environment, even though it's running not run on NCI though..

This also points to a wider issue with tests - there's a number of ACCESS-OM2 specific ones that test that some paths in `config.yaml` match a path in `/g/data/vk83`. I think for the ACCESS-OM2 configs, they are specifically meant to run on NCI at the moment. If payu could be run on other supercomputers, ACCESS-OM2 configs at least will need some updates/or separate configurations. The supercomputer specific tests might need to be separated, possible using markers for NCI/gadi. Or if the released modules and scripts had the same relative directory structure, could have an generic argument for the "release" directory, e.g. `/g/data/vk83/` in gadi's case?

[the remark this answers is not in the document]
jo-basevi: This could moved to a future issue if testing on other supercomputers becomes a requirement

src/model_config_tests/test_config.py
CodeGat: Shelving this bit for now, I think.

CodeGat (approved):

PR #31, opened 2024-07-10
Fix `configs-ci` `!bump` command

CodeGat:
See failed run: https://github.com/ACCESS-NRI/access-om2-configs/actions/runs/9866119810/job/27244325155#step:4:10

This was a regression from moving the configs from `access-om2-configs` into `model-config-tests`. 

In this PR:
* Add back the `env.GH_TOKEN` to the `gh` run command.

PR #32, opened 2024-07-10
Update other tokens used for `gh` commands

CodeGat:
A similar regression that was found in https://github.com/ACCESS-NRI/model-config-tests/pull/31
See failed run: https://github.com/ACCESS-NRI/access-om2-configs/actions/runs/9866828779/job/27246236631

In this PR:
* Add back the `env.GH_TOKEN` when using `gh` commands


CodeGat:
Similarly, I will merge this one as there is no one else to review at this time, and this needs to be fixed. Will ask for a reviewer so it can be done after.

harshula (commented):
Reviewed in VC.

Issue #33, opened 2024-07-10
`config ci`: Features/Fixes to Help UX


CodeGat:
Now that devs (@dougiesquire :eyes:)  are using the workflow, there are some features being requested (and I'll add a couple more nice-to-haves from my own perspective, too). This issue can be a general dumping ground for features until we create a Zenhub Epic. 

- [x] Have links to both the compared checksum directory (will be a hyperlink to the GitHub repo in `testing/checksum` at the compared Config Tag) and the current PR checksum directory (at that point will be an artifact, so will be a download link and a `cd testing/checksum` directive) in the [Bitwise Reproducibility](https://github.com/ACCESS-NRI/access-om2-configs/pull/126#issuecomment-2218962311) comment. See https://github.com/ACCESS-NRI/model-config-tests/pull/34
- [x] Rather than needing to rebase the `dev-*` onto `release-*` every time we have a new `release-*` config, search for the previous `release-*` config tag on the `${{ github.base_ref }}` rather than the `${{ github.head_ref }}`. Since we have a well-defined naming scheme, it should be easy to checkout. See https://github.com/ACCESS-NRI/model-config-tests/pull/35
- [x] Investigate the suitability of `dev-*` branches at all - we seem to immediately merge into `release-*` anyway. If we need to have somewhere to develop and work on configurations, why not just use the PR branch? Look into previous discussions for the rationale. See https://github.com/ACCESS-NRI/model-config-tests/issues/38 and a future issue.
- [x] Consider removing the `Create Initial Checksum` workflow for the initial development of checksums. The workflow actually gets quite far along without a checksum to compare to, so it might be worth just allowing people to merge to the `release-*` branch without a checksum to compare against. See https://github.com/ACCESS-NRI/model-config-tests/pull/36

Please feel free to comment on this issue for further improvements that can be made to the workflow.

CodeGat:
Closing this as tasks have either been turned into their own issues, or completed.

PR #36, opened 2024-07-17
Remove `generate-checksums.yml` Workflow

CodeGat:
Before merging, determine if there are any use cases for a user-dispatchable generator of checksums still. Maybe ask @aidanheerdegen. 

In this PR:
* Remove the `generate-checksums.yml` workflow (`Create Initial Checksums`) as the current `config-pr-*.yml` workflow as it stands will commit the checksums whether a previous comparable checksum exists. 

Requires #35 (changes to `test-repro`, etc)
References #33


CodeGat:
There still seems to be use cases (especially in `OM3`-land) that need this functionality.

PR #41, opened 2024-08-06
`ACCESS-ESM1.5`: QA Tests

### LINKED ISSUE #21: ACCESS-ESM1.5 testing
As part of the [ACCESS-ESM1.5 Release](https://forum.access-hive.org.au/t/access-esm1-5-access-nri-flagship-release-plans/1772) there needs to be reproducibility and QA tests for the [ACCESS-ESM1.5 model configurations](https://github.com/ACCESS-NRI/access-esm1.5-configs/).

As ACCESS-ESM1.5 uses MOM5 the same reproducibility tests should suffice in the first instance. This code could be moved to a `MOM5` class and then used by ACCESS-OM2 and ACCESS-ESM1.5.

CodeGat:
Add QA tests for `ACCESS-ESM1.5` to the `model-config-tests` package. Requires bump to `0.0.3` when merged. 

In this PR:
* Move all `test_*_config.py` files into new `qa` directory
* Added `test_access_esm1p5_config.py` -  see https://github.com/ACCESS-NRI/model-config-tests/issues/21#issuecomment-2270383799
* Added `access_esm1p5` QA pytest marker
* Bumped version to `0.0.3`

Closes #21 

Pinging @jo-basevi so she can follow along on my progress :smile:


CodeGat:
Note `access-esm1.5-configs` will require a (not yet released `0.0.3` `model-config-tests` package to function

CodeGat:
Might need a bit of a sanity check on the use of parameterization in the pytests.
The code coverage is also failing...Do I need to make tests for my tests...?

CodeGat:
Tested this on Gadi with https://github.com/ACCESS-NRI/access-esm1.5-configs/commit/9191fd9d57a5b67976b41e0dc2e01d3543f37a46 `pre-industrial` configuration. A bunch of errors because no `metadata.yaml`, and the failures seem expected. Might need to check the relevance of the 'configs' test marker tests...
Also still don't know what to do with the code coverage for the tests themselves...

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -0,0 +1,167 @@
...
+
+### Bunch of expected values for tests ###
+VALID_REALMS: set[str] = {"atmosphere", "land", "ocean", "ocnBgchm", "seaIce"}
+VALID_KEYWORDS: set[str] = {"global", "access-esm1.5"}
+VALID_NOMINAL_RESOLUTION: str = "100 km"
+VALID_REFERENCE: str = "https://doi.org/10.1071/ES19035"
+VALID_PREINDUSTRIAL_START: dict[str, int] = {"year": 101, "month": 1, "day": 1}
+VALID_HISTORICAL_START: dict[str, int] = {"year": 1850, "month": 1, "day": 1}
+VALID_RUNTIME: dict[str, int] = {"year": 1, "month": 0, "day": 0}
+VALID_RESTART_FREQ: str = "20YS"
```
aidanheerdegen, line 22: @blimlim talked to Martin and settled on "10YS" as a standard for these configs. Sorry for the bum steer on that one.

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -0,0 +1,167 @@
...
+        self.config_scenario = self.set_config_scenario()
+        self.config_modifiers = self.set_config_modifiers()
+
+    def set_config_scenario(self) -> str:
+        # Regex below is split into three sections:
+        # Config type start section: '^.+-' for 'release-', 'dev-'...
+        # Scenario section: '([^+]+)' for 'preindustrial', 'historical'...anything that isn't the '+' modifier sigil
+        # Modifiers end section: '(?:\+.+)*' any amount of '+modifer' sections
+        scenario_match = re.match(
+            r"^.+-(?P<scenario>[^+]+)(?:\+.+)*$", self.branch_name
```
aidanheerdegen, line 49: Should this be more explicit and require either `release` or `dev` before the `-` separator? I kinda think we should. We're not promising to deal with anything other than that are we?

Also do we care if it is `release` or `dev`? I guess not if we're not saving it in the state, just curious.

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -0,0 +1,167 @@
...
+            pytest.fail(
+                f"Could not find a scenario in the branch {self.branch_name}. "
+                + "Branches must be of the form 'type-scenario[+modifier...]'. "
+                + "See README.md for more information."
+            )
+        return scenario_match.group("scenario")
+
+    def set_config_modifiers(self) -> list[str]:
+        # Regex below is essentially 'give me the 'modifier' part in all the '+modifier's in the branch name'
+        return re.findall(r"\+([^+]+)", self.branch_name)
```
aidanheerdegen, line 61: Yeah tests of functions like this would be kinda nice.

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -0,0 +1,167 @@
...
+        # Default to current branch name
+        branch_name = get_git_branch_name(control_path)
+        assert (
+            branch_name is not None
+        ), f"Failed getting git branch name of control path: {control_path}"
+        warnings.warn(
+            "Target branch is not specifed, defaulting to current git branch: "
+            f"{branch_name}. As some ACCESS-ESM1.5 tests infer information, "
+            "such as resolution, from the target branch name, some tests may "
+            "not be run. To set use --target-branch flag in pytest call"
```
aidanheerdegen, line 77: It isn't the case that we infer resolution from branch name for ESM configs. We may infer other things, but I think generally we'd be using the scenario and modifiers to determine optional tests.

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -0,0 +1,167 @@
...
+        # TODO: We don't check for high resolution here like we do in the ACCESS-OM2 version of the test. Should we?
+        pattern = r"/g/data/vk83/apps/mppnccombine-fast/.*/bin/mppnccombine-fast"
+        if "collate" in config:
+            assert re.match(
+                pattern, config["collate"]["exe"]
+            ), "Expect collate executable set to mppnccombine-fast"
+
+            assert config["collate"][
+                "mpi"
+            ], "Expect `mpi: true` when using mppnccombine-fast"
```
aidanheerdegen, line 167: Oh bugger. Just realised we shouldn't be testing `mppnccombine-fast`. We want to use the `mppnccombine` build by `spack` which looks like this

https://github.com/ACCESS-NRI/model-config-tests/blob/7053608e57c64864fe9f1b135e472eb384435e04/tests/resources/access-om2/configurations/release-1deg_jra55_ryf/config.yaml#L77C8-L77C26

aidanheerdegen (changes requested):
Soz, a couple of changes and some questions/comments.

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Good shout on the regex. Will update. 
I can't think of any reason for QA tests to treat them any differently whether they're `release-*` or `dev-*`, so I don't think we need to capture that information.

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Updated in https://github.com/ACCESS-NRI/model-config-tests/pull/41/commits/c4d8a7d130b1a3764c34ba3c7cd0cbd7cb7971f8

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Updated in https://github.com/ACCESS-NRI/model-config-tests/pull/41/commits/c4d8a7d130b1a3764c34ba3c7cd0cbd7cb7971f8

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Updated in https://github.com/ACCESS-NRI/model-config-tests/pull/41/commits/c4d8a7d130b1a3764c34ba3c7cd0cbd7cb7971f8

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Updated in https://github.com/ACCESS-NRI/model-config-tests/pull/41/commits/c4d8a7d130b1a3764c34ba3c7cd0cbd7cb7971f8

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: And in https://github.com/ACCESS-NRI/model-config-tests/pull/41/commits/8f6ff1e88ec3f5872757e78867c964d0625c02d0 because I had a strange `?` after `(?:release|dev)`

aidanheerdegen (approved):
LGTM!

aidanheerdegen (approved):
I'll approve, but if you think we should make the changes to the module use path now, then feel free to do so and ask for yet another review

https://github.com/ACCESS-NRI/model-config-tests/blob/bdfb85f19e5c4f6d20101af4906a97aa2ed2bb3b/src/model_config_tests/qa/test_config.py#L26

aidanheerdegen (dismissed):
LG(reat)TM

aidanheerdegen (approved):
Nice catch.

PR #43, opened 2024-08-12
Fixes for `ACCESS-ESM1.5` Config Tests

CodeGat:
Through testing [actual ESM1.5 configs](https://github.com/ACCESS-NRI/access-esm1.5-configs/pull/44/checks?check_run_id=28633473681) we have found some errors in the `VALID_*_START` and `VALID_RUNTIME` constants. Also added a check for nonexistence of `collate.exe.mpi`.

Pinging @blimlim since he helped out correcting these errors.


src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -24,9 +24,9 @@
 VALID_KEYWORDS: set[str] = {"global", "access-esm1.5"}
 VALID_NOMINAL_RESOLUTION: str = "100 km"
 VALID_REFERENCE: str = "https://doi.org/10.1071/ES19035"
-VALID_PREINDUSTRIAL_START: dict[str, int] = {"year": 101, "month": 1, "day": 1}
-VALID_HISTORICAL_START: dict[str, int] = {"year": 1850, "month": 1, "day": 1}
-VALID_RUNTIME: dict[str, int] = {"year": 1, "month": 0, "day": 0}
+VALID_PREINDUSTRIAL_START: dict[str, int] = {"year": 101, "month": 1, "days": 1}
+VALID_HISTORICAL_START: dict[str, int] = {"year": 1850, "month": 1, "days": 1}
+VALID_RUNTIME: dict[str, int] = {"years": 1, "months": 0, "days": 0}
```
truth-quark, line 29: Do these fields need to be plural only OR plural & singular?

src/model_config_tests/qa/test_access_esm1p5_config.py
```diff
@@ -181,6 +181,10 @@ def test_mppnccombine_fast_collate_exe(self, config):
                 "collate.exe", "config.yaml", VALID_MPPNCCOMBINE_EXE
             )
 
+            assert "mpi" in config["collate"], error_field_nonexistence(
```
truth-quark, line 184: `codecov` in GitHub is warning about this missing test coverage. I this something required for the PR?

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: Plural only for `VALID_RUNTIME`. The updated way is the way that Payu understands it. It's not internally consistent, but that is a pull request for Payu first.

src/model_config_tests/qa/test_access_esm1p5_config.py
CodeGat: This would require adding tests for tests, which I don't think we have time for at the moement. It's on the list, though: https://github.com/ACCESS-NRI/model-config-tests/issues/44

truth-quark: Ok, that sounds like an endless self referencing problem...

truth-quark (approved):
Looks good to go!

PR #45, opened 2024-08-12
`ACCESS-ESM1.5` Further Test Fixes

CodeGat:
After testing the `metadata.yaml` tests on an actual ESM1.5 configuration, it was found that more expected fields were actually incorrect. 

In this PR:
* Change 'ocnBgchm' to 'ocnBgchem'
* Change 'keyword' to 'keywords'
* Update version of `model-config-tests` to `0.0.5`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "model_config_tests"
-version = "0.0.4"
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM

PR #47, opened 2024-08-13
`ACCESS-ESM1.5` `mppnccombine` MPI Test Fix

CodeGat:
See https://github.com/ACCESS-NRI/access-esm1.5-configs/pull/44/checks?check_run_id=28688143378

We allow `collate.mpi` to be either `true` or `false`, rather than only `true`. 

Since converting the `.yaml` file into a Python `dict` also converts the above to a `bool`, test that the resulting value is an instance of `bool`.

In this PR:
* Update the test
* Update the version of `model-config-tests` to `0.0.7`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "model_config_tests"
-version = "0.0.6"
+version = "0.0.7"
 authors = [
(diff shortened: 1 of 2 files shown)


harshula (approved):
Reviewed in-person.

PR #49, opened 2024-08-14
Workflow-verified Commits

### LINKED ISSUE #48: Allow Workflow-Authoured Verified Commits
Currently, we use `access-bot` for our workflow-authored commits, but we don't use the associated signing key to make the commits show up as verified.

See https://github.com/ACCESS-NRI/dev-docs/wiki/Continuous-integration-and-Deployment#verified-commits-from-a-workflow for the solution.

Similar to https://github.com/ACCESS-NRI/build-cd/issues/115

CodeGat:
In this PR:
* Import commit-signing keys using org secrets

Closes #48


jo-basevi (approved):
LGTM

PR #55, opened 2024-08-26
Add `git` Tag Message

### LINKED ISSUE #53: Config versions bump failing due to annotated tags
The workflow to bump tags in the ACCESS-ESM1.5 configs is failing

https://github.com/ACCESS-NRI/access-esm1.5-configs/actions/runs/10537864862/job/29199729736#step:5:6

https://github.com/ACCESS-NRI/access-esm1.5-configs/actions/runs/10523427873/job/29158089899#step:5:6

The error occurs when the new tag is being created
```
Run git tag release-historical+concentrations-1.0
  git tag release-historical+concentrations-1.0
  git push --tags
  shell: /usr/bin/bash -e {0}
```

the error is:
```
error: Terminal is dumb, but EDITOR unset
Please supply the message using either -m or -F option.
Error: Process completed with exit code 1.
```

So it seems like the `git tag` is trying to created annotated tags, and it is trying to open an editor to create a commit message and failing.

I think it is related to this change

https://github.com/ACCESS-NRI/model-config-tests/pull/49

so the bot now signs commits.

According to the [git tag docs](https://git-scm.com/book/en/v2/Git-Basics-Tagging) `git` defaults to lightweight commits only when a number of other options *aren't* defined:

> Lightweight Tags
Another way to tag commits is with a lightweight tag. This is basically the commit checksum stored in a file — no other information is kept. To create a lightweight tag, don’t supply any of the -a, -s, or -m options, just provide a tag name

Seeing as we're now effectively specifying `-s`, we no longer have lightweight commits. 

As far as a solution goes, as there is no way to force lightweight commits, the CI will have to add a commit message using the `-m` command line option. It could be an empty string, but might as well be something useful identifying a version bump and the workflow that did it?

CodeGat:
In this PR:
* Added a `-m` flag to the `git tag` call as signed commits require a tag message

Closes #53

### DIFF
--- a/.github/workflows/config-pr-3-bump-tag.yml
+++ b/.github/workflows/config-pr-3-bump-tag.yml
@@ -56,7 +56,7 @@ jobs:
           TAG: ${{ steps.tag.outputs.version }}
           STATUS: ${{ endsWith(steps.tag.outputs.version, '.0') && 'breaking' || 'preserving' }}
         run: |
-          git tag ${{ env.TAG }} -m "Repro-${{ env.STATUS }} update to ${{ github.ref_name }} via model-config-tests '${{ github.workflow }}' workflow."
+          git tag ${{ env.TAG }} -m "Repro-${{ env.STATUS }} update to ${{ github.ref_name }} via model-config-tests 'config-pr-3-bump-tag.yml' workflow."
           git push --tags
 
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
Looks good to me!

PR #65, opened 2024-09-10
Dispatchable Workflow for Service User Config Inputs Copying

### LINKED ISSUE #64: Create User-Dispatchable Workflow to Move Config Inputs
Currently, we require someone to login as `tm70_ci` to move configuration inputs into `/g/data/vk83/configurations/inputs` since it is restricted. 

This is a bit painful, so create a workflow that takes in:

* Remote file source
* Remote file target
* Whether to overwrite the target with the source
* The ACLs that the file will have (create a sensible default)

Require sign off by developers, and log the inputs so we have visibility. 

The workflow should copy source to target as `tm70_ci` and set appropriate ACLs.

CodeGat:
## Background

This PR adds a new user-dispatchable workflow for the moving of configuration input files across the same remote host as a service user. This is useful for allowing `/g/data/vk83` to remain as a kind of human-free-zone when it comes to important files.

## Questions for Reviewers

* Is the given default `target-acl-spec` correct?

## The PR 

In this PR:
- **Add new inputs-remote-copy dispatchable workflow**
- **README.md: Add info on inputs-remote-copy.yml**

Repository Settings Modified:
- Added `Gadi` GitHub Environment for access to remote `Gadi`

Closes ACCESS-NRI/model-config-inputs#1

### DIFF
--- a/.github/workflows/inputs-remote-copy.yml
+++ b/.github/workflows/inputs-remote-copy.yml
@@ -25,6 +25,7 @@ on:
         type: string
         required: true
         # Default to no write for everyone except tm70_ci
+        # TODO: This default will probably not work for other `remote-environment`s
         default: >-
           u::rwx,
           u:tm70_ci:rwx,
--- a/.github/workflows/inputs-remote-copy.yml
+++ b/.github/workflows/inputs-remote-copy.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Verify Remote Target
         run: |
-          if [[ "${{ contains(inputs.target, '/g/data/vk83/configurations/inputs') }}" == "true" ]]; then
+          if [[ "${{ contains(inputs.target, vars.CONFIGS_INPUT_DIR) }}" == "true" ]]; then
             echo "::error::Remote target '${{ inputs.target }}' doesn't look like a configurations input directory."
             exit 1
           fi
--- a/.github/workflows/inputs-remote-copy.yml
+++ b/.github/workflows/inputs-remote-copy.yml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Verify Remote Target
         run: |
-          if [[ "${{ contains(inputs.target, vars.CONFIGS_INPUT_DIR) }}" == "true" ]]; then
+          if [[ "${{ startsWith(inputs.target, vars.CONFIGS_INPUT_DIR) }}" == "true" ]]; then
             echo "::error::Remote target '${{ inputs.target }}' doesn't look like a configurations input directory."
             exit 1
           fi
--- a/.github/workflows/inputs-remote-copy.yml
+++ b/.github/workflows/inputs-remote-copy.yml
@@ -1,0 +1,89 @@
+name: Config Inputs Remote Copy
+run-name: "${{ inputs.remote-environment }} Config Input Copy to ${{ inputs.target }}"
+on:
+  workflow_dispatch:
+    inputs:
+      remote-environment:
+        type: choice
+        required: true
+        description: The Github Environment for the given remote
+        options:
+          - Gadi
+      source:
+        type: string
+        required: true
+        description: Remote absolute path to configuration input file/folder
+      target:
+        type: string
+        required: true
+        description: Remote absolute path for the copied configuration input file/folder
+      overwrite-target:
+        type: boolean
+        required: true
+        description: Overwrite the remote target if it already exists
+      target-acl-spec:
+        type: string
+        required: true
+        # Default to no write for everyone except tm70_ci
+        default: >-
+          u::rwx,
+          u:tm70_ci:rwx,
+          g::r-x,
+          m::rwx,
+          o::---,
+          d:u::rwx,
+          d:u:tm70_ci:rwx,
+          d:g::r-x,
+          d:m::rwx,
+          d:o::---
+        description: ACL spec to be passed to `setfacl -m` for the given target
+jobs:
+  setup:
+    name: Setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log inputs
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/inputs-remote-copy.yml
```diff
@@ -0,0 +1,90 @@
...
+        options:
+          - Gadi
+      source:
+        type: string
+        required: true
+        description: Remote absolute path to configuration input file/folder
+      target:
+        type: string
+        required: true
+        description: Remote absolute path for the copied configuration input file/folder
```
aidanheerdegen, line 19: We *could* (not saying this is a good idea) split the path into a root and a final calVer sub-dir, and provide a default for the latter to help users who aren't sure how to version inputs.

.github/workflows/inputs-remote-copy.yml
```diff
@@ -0,0 +1,90 @@
...
+      remote-environment:
+        type: choice
+        required: true
+        description: The Github Environment for the given remote
+        options:
+          - Gadi
+      source:
+        type: string
+        required: true
+        description: Remote absolute path to configuration input file/folder
```
aidanheerdegen, line 15: We need to be quite careful with trailing slashes in rsync

https://stackoverflow.com/a/31278462

It is *really* easy to not do what you expected.

aidanheerdegen (changes requested):
Looks great, thanks! 

I think we should add some logging, so we have a record of what was changed and when. Logs should be in the `inputs` dir and world readable, but not writeable natch. Also machine-readable. CSV is fine if nothing else springs to mind.

Thanks!

.github/workflows/inputs-remote-copy.yml
```diff
@@ -0,0 +1,90 @@
...
+          u:tm70_ci:rwx,
+          g::r-x,
+          m::rwx,
+          o::---,
+          d:u::rwx,
+          d:u:tm70_ci:rwx,
+          d:g::r-x,
+          d:m::rwx,
+          d:o::---
+        description: ACL spec to be passed to `setfacl -m` for the given target
```
CodeGat, line 40: @aidanheerdegen how does this default ACLs look?

.github/workflows/inputs-remote-copy.yml
CodeGat: This is true. But I think that it should be a responsibility of the user to write their `inputs.source` correctly rather than the CI to modify the command that they have explicitly written. What do you think?

aidanheerdegen: Depends what this workflow is designed for. We could make not use `rsync` at all and be very prescriptive (and helpful!) and automate versions etc and be designed to just put new files in a single location to update an existing large directory tree.

Then we could have a different, related, workflow that is really supposed to copy whole directory trees, but it would happen rarely and maybe would definitely *not* overwrite any existing files.

If we don't do above we should add some warnings in the description/docs to be mindful of trailing slashes!

.github/workflows/inputs-remote-copy.yml
aidanheerdegen: They seem fine. I haven't tested them in real world though.

.github/workflows/inputs-remote-copy.yml
```diff
@@ -0,0 +1,90 @@
...
+            exit 1
+          fi
+
+      - name: Verify Valid ACL Spec
+        if: inputs.target-acl-spec != ''
+        env:
+          TEST_DIR: ./test
+        run: |
+          mkdir ${{ env.TEST_DIR }}
+          setfacl --test --recursive --modify "${{ inputs.target-acl-spec }}" ${{ env.TEST_DIR }}
```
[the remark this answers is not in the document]
CodeGat: It doesn't...and it also doesn't let you set ACLs for users that don't exist - just tested it on my own personal laptop. Even though the runners run passwordless sudo, I'm not too sure if it'd be a good idea to start creating users and groups just to test this stuff out...but I will see on my test organisation if it's possible

.github/workflows/inputs-remote-copy.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+      - name: Log inputs
+        run: |
+          echo ""::notice::Copy on ${{ inputs.remote-environment }} from '${{ inputs.source }}' to '${{ inputs.target }}' with ACLs '${{ inputs.target-acl-spec }}'"
+          echo "::${{ inputs.overwrite-target && 'warning' || 'notice' }}::This operation ${{ inputs.overwrite-target && 'WILL' || 'will not' }} overwrite ${{ inputs.target }}"
+
+      - name: Verify Remote Target
+        run: |
+          if [[ "${{ contains(inputs.target, vars.CONFIGS_INPUT_DIR) }}" == "true" ]]; then
+            echo "::error::Remote target '${{ inputs.target }}' doesn't look like a configurations input directory."
```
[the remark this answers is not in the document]
CodeGat: Done in https://github.com/ACCESS-NRI/model-config-tests/pull/65/commits/230a6bcc1cb6e6ec22dbd12f3b81aa9e11961b6a

CodeGat:
Closing this PR as we will be moving it from this repository to https://github.com/ACCESS-NRI/model-config-inputs/. All feedback from this PR will be addressed in a new PR, and will be linked here when created.

.github/workflows/inputs-remote-copy.yml
CodeGat: I think for now we should go with the "warning the user" bit. We can iterate upon this idea in the future, I think.

.github/workflows/inputs-remote-copy.yml
CodeGat: I think in the first instance, we should start with explicit commands (as that is what users were originally doing (`cp` their work into the input dir)), and add helpful stuff incrementally

PR #71, opened 2024-10-03
Sanitize Dispatch Inputs

CodeGat:
This PR guards against the possibility of accidentally-spaced user input in path-like things - see https://github.com/ACCESS-NRI/access-om3-configs/actions/runs/11137950615/job/30952195151#step:3:6

In this PR:
- **Quoting path parameters**
- **Added step to check that certain environment vars exist**


jo-basevi (approved):
Changes look good to me!

Issue #72, opened 2024-10-03
Merge `generate-checksums.yml` into `test-repro.yml`


CodeGat:
`generate-checksums.yml` is essentially the same code as in `test-repro.yml`. We shouldn't have to modify two of the same file, essentially. 

Incorporate `generate-checksums.yml` into `test-repro.yml`.

Issue #68, opened 2024-10-02
Incorporate `check-checksum` into `test-repro.yml`


CodeGat:
Currently, there is a common `check-checksum` job run after `repro-ci` in both `config-pr-1-ci.yml` and `config-schedule-2-start.yml` which downloads the test artifact and adds the test results to the given PR.

Instead of doing the same job twice, incorporate it into the `test-repro.yml` reusable workflow. If it is a possibility that someone might call this workflow while not `on.pull_request`, make the `check-checksum` part only run `if: github.event_name == 'pull_request'`.

PR #73, opened 2024-10-07
`!test` Comment Command

### LINKED ISSUE #38: Add Repro Checks to `dev-*` branches
## Background

Since the QA checks are run before the Repro checks in `release-*` branches, we might as well add them to `dev-*` as well. 
This gives the `dev-*` PRs the ability to test reproducibility before opening a `release-*` PR and potentially having to go back and open a `dev-*` PR again if one gets an unfavourable reproducibility result.  

## Considerations
* We will have to remove the requirement of Environments only being able to deploy to protected branches, as the it takes the `source` branch rather than the `target` branch.

CodeGat:
Add a `!test repro [commit]` command to allow commenters to commit and compare repro runs that aren't on `release-*` branches. 

In this PR:
- **Add config-comment-repro.yml workflow for repro tests on any branch!**
- **Added ability to commit the result of the repro test to the PR**

Closes #38

### DIFF
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Get test type
         id: test
         run: |
-          command="${{ github.event.issue.comment.body }}"
+          command="${{ github.event.comment.body }}"
           read -ra command_tokens <<< "$command"
           type_in_comment="${command_tokens[1]}"
           type_in_comment_valid=false
@@ -99,7 +99,7 @@ jobs:
           done
 
           if [[ "$type_in_comment_valid" == "false" ]]; then
-            echo "::error::The command '${{ github.event.issue.comment.body }}' doesn't have a valid TYPE. Was given '$type_in_command', but needed to be one of: ${{ env.USAGE_TYPES }}."
+            echo "::error::The command '${{ github.event.comment.body }}' doesn't have a valid TYPE. Was given '$type_in_command', but needed to be one of: ${{ env.USAGE_TYPES }}."
             exit 1
           fi
 
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -12,8 +12,40 @@ env:
   USAGE_TYPES: repro
   RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
 jobs:
+  permission-check:
+    name: Permission Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Determine if commenter has permission to run the command
+        id: commenter
+        # TODO: The permissions-checking section seems ripe for turning into an action!
+        run: |
+          commenter_permissions=$(gh api \
+              repos/${{ github.repository }}/collaborators/${{ github.event.comment.user.login }}/permission \
+              --jq '.permission'
+          )
+          if [[ "$commenter_permission" == "admin" || "$commenter_permission" == "write" ]]; then
+            echo "Commenter has at least write permission"
+            echo "permission=true" >> $GITHUB_OUTPUT
+          else
+            echo "Commenter does not have at least write permission"
+            echo "permission=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: React to '!test'
+        uses: access-nri/actions/.github/actions/react-to-comment@main
+        with:
+          reaction: ${{ steps.commenter.outputs.permission == 'true' && 'rocket' || 'confused' }}
+          token: ${{ github.token }}
+
+      - name: Fail if no permissions
+        if: steps.commenter.outputs.permission == 'false'
+        run: exit 1
+
   ci-config:
     name: Read CI Testing Configuration
+    needs:
+      - permission-check
     runs-on: ubuntu-latest
     outputs:
       markers: ${{ steps.repro-config.outputs.markers }}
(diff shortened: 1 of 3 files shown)


.github/workflows/config-comment-test.yml
```diff
@@ -0,0 +1,366 @@
...
+        with:
+          comment: |
+            ${{ env.RESULT }}
+
+            When comparing:
+
+            - `${{ needs.prepare-command.outputs.config-ref }}` (checksums created using commit ${{ env.CONFIG_REF_URL }}), against
+            - `${{ needs.prepare-command.outputs.compared-config-ref }}` (checksums in commit ${{ env.COMPARED_CONFIG_REF_URL }})
+
+            ${{ (needs.prepare-command.outputs.commit-requested == 'true' && steps.results.outputs.result == 'fail') && ':wrench: The checksums will be committed to this PR, as they differ.' || '' }}
```
aidanheerdegen, line 279: It doesn't report that the checksums differ if `commit-requrest == false`. I think it should report this separately, and then check for `commit-requested` and `result` and report if they'll be committed.

.github/workflows/config-comment-test.yml
```diff
@@ -0,0 +1,366 @@
...
+
+            The experiment can be found on Gadi at `${{ needs.repro.outputs.experiment-location }}`, and the test results at ${{ steps.results.outputs.check-run-url }}.
+
+            The checksums generated by this `!test` command are found in the `testing/checksum` directory of ${{ needs.repro.outputs.artifact-url }}.
+
+            The checksums compared against are found here ${{ github.server_url }}/${{ github.repository }}/tree/${{ needs.prepare-command.outputs.compared-config-hash }}/testing/checksum
+
+            </details>
+
+  commit:
```
aidanheerdegen, line 292: Just a general comment that this has a fair bit in common with https://github.com/ACCESS-NRI/model-config-tests/blob/main/.github/workflows/config-pr-2-confirm.yml#L97-L161

which I am sure you know. I can see there are differences, but code duplication ugh.

aidanheerdegen (changes requested):
Nice work, thanks! 

I note that "just add a the ability to run repro test through ChatOps" ends up being a lot more complicated than I realised. Sorry about that!

.github/workflows/config-comment-test.yml
CodeGat: Yeah, I have a fair bit of code-deduplication to do: https://github.com/ACCESS-NRI/model-config-tests/issues/72 and https://github.com/ACCESS-NRI/model-config-tests/issues/68 - this will be my next PR I think!

.github/workflows/config-comment-test.yml
```diff
@@ -0,0 +1,366 @@
...
+          fetch-tags: true
+
+      - name: Get test type
+        id: test
+        run: |
+          command="${{ github.event.comment.body }}"
+          read -ra command_tokens <<< "$command"
+          type_in_comment="${command_tokens[1]}"
+          type_in_comment_valid=false
+
```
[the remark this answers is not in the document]
CodeGat: It isn't - it's checking the that the `TYPE` of the `!test` is one that we support (currently just `!test repro`, but we may have `!test qa` or `!test performance` for example). 
But this has pointed to the fact that we don't actually verify the name of the command in this workflow (only outside of it), so I've added that. See https://github.com/ACCESS-NRI/model-config-tests/pull/73/commits/f650189b87ba9c2068f4159968b8c5d7363f4b0d

.github/workflows/config-comment-test.yml
```diff
@@ -0,0 +1,366 @@
...
+          for type in ${{ env.USAGE_TYPES }}; do
+            if [[ "$type_in_comment" == "$type" ]]; then
```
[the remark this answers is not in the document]
CodeGat: Done in https://github.com/ACCESS-NRI/model-config-tests/pull/73/commits/98de64b2eb5b3af774b601d8a351a479e3bf0eba

README.md
```diff
@@ -148,3 +147,33 @@ Using it has some requirements outside of just filling in the inputs: One must h
...
+
+```txt
+D   (PR Target Branch)
+|
+| C (PR Source Branch)
+| |
+| B
+|/
+A   (Common Ancestor)
+```
```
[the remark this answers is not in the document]
CodeGat: Done in https://github.com/ACCESS-NRI/model-config-tests/pull/73/commits/98de64b2eb5b3af774b601d8a351a479e3bf0eba

.github/workflows/config-comment-test.yml
```diff
@@ -119,7 +119,16 @@ jobs:
...
           fetch-tags: true
 
-      - name: Get test type
+      - name: Verify !test
+        run: |
+          if [[ "${{ startsWith(github.event.comment.body, '!test') }}" == "true" ]]; then
+            echo 'Command starts with !test'
+          else
+            echo "::error::Usage: ${{ env.USAGE }}"
+            echo "::error::Command must start with !test to invoke model-config-tests' config-comment-test.yml."
```
aidanheerdegen, line 128: What happens if someone does `!bump`? Are the compatible?

.github/workflows/config-comment-test.yml
CodeGat: It does report the differing checksums in the `env.RESULT` part - https://github.com/ACCESS-NRI/model-config-tests/pull/73/files#diff-3030a45e2717eb6aff0b15ab7e53158851b49ea677462370d257599c221e8925R273-R275

.github/workflows/config-comment-test.yml
CodeGat: The `!bump` command is handled in https://github.com/ACCESS-NRI/model-config-tests/blob/main/.github/workflows/config-pr-2-confirm.yml, not here. Maybe in future we could consolidate all the comment commands, but that would be a very very large file :D

aidanheerdegen (approved):
Nice work!

PR #76, opened 2024-10-17
Merge common `check-checksum` job into `test-repro.yml`

CodeGat:
We've been amassing a series of jobs that run after `repro-ci` that all do the same thing: download the testing artifact, parsing the test data, and returning the result. This has become so common that it should be a part of `test-repro.yml` itself. 

In this PR:
- **test-repro.yml: Move `check-checksum` into workflow**
- **config-schedule-2-start.yml: Move `check-checksum` out**
- **config-pr-1-ci.yml: Move `check-checksum` out**
- **config-comment-test.yml: Move `check-checksum` out**

Closes #68

PR #77, opened 2024-10-18
Merge Common `test-repro` Functionality in `generate-checksums.yml`

### LINKED ISSUE #72: Merge `generate-checksums.yml` into `test-repro.yml`

CodeGat:
In this PR, we remove the `generate-checksums` job from `generate-checksums.yml` in favour of calling the perfectly good `test-repro.yml` workflow, removing further duplication from the workflows in the code. 

Closes #72

### DIFF
--- a/.github/workflows/generate-checksums.yml
+++ b/.github/workflows/generate-checksums.yml
@@ -1,4 +1,5 @@
 name: Initial Checksums
+run-name: Initial Checksums for ${{ inputs.config-branch-name }}
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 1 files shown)


CodeGat:
Note, will have to check that we don't have to add an input to `test-repro` that disables the `check` run from a `on.workflow_dispatch` workflow...
We may have to add `permissions.checks:write` once https://github.com/ACCESS-NRI/model-config-tests/pull/76 is merged

.github/workflows/generate-checksums.yml
```diff
@@ -178,13 +72,13 @@ jobs:
...
-          name: ${{ needs.generate-checksum.outputs.artifact-name }}
-          path: ${{ env.OUTPUT_LOCAL_LOCATION }}
+          name: ${{ needs.repro-ci.outputs.artifact-name }}
+          path: ${{ env.ARTIFACT_LOCAL_LOCATION }}
 
       - name: Move Checksums to Repo
         run: |
           mkdir -p ${{ inputs.committed-checksum-location }}
-          mv ${{ env.OUTPUT_LOCAL_LOCATION }}/checksum/* ${{ inputs.committed-checksum-location }}
+          cp --recursive --verbose ${{ env.ARTIFACT_LOCAL_LOCATION }}/*/* ${{ inputs.committed-checksum-location }}
```
aidanheerdegen, line 81: This has switched from `mv` to `cp --recursive` and including everything in a sub-directory of `env.ARTIFACT_LOCAL_LOCATION` rather than just under the `checksum` sub-dir.

Can you explain those changes so we have a record of why? Thanks!

.github/workflows/generate-checksums.yml
```diff
@@ -214,8 +108,10 @@ jobs:
 
       - name: Tag Checksums in Repo
         if: inputs.committed-checksum-tag != ''
+        env:
+          STATUS: ${{ endsWith(inputs.committed-checksum-tag, '.0') && 'breaking' || 'preserving' }}
         run: |
-          git tag ${{ inputs.committed-checksum-tag }}
+          git tag ${{ inputs.committed-checksum-tag }} -m "Repro-${{ env.STATUS }} update to ${{ inputs.config-branch-name }} via model-config-tests 'generate-checksums.yml' workflow."
```
aidanheerdegen, line 114: I like this informative comment on the git tag.

aidanheerdegen (approved):
Stripping out a big chunk of dupe'd code. Nice. Just a couple of commments.

.github/workflows/generate-checksums.yml
CodeGat: This will support the case where we have potentially other subdirectories that we want to copy, like `additional-artifact-content-paths`, or maybe `performance` or something like that.

tmcadam (commented):
LGTM Good to de-dupe!!

tmcadam:
Looks good!

tmcadam (commented):
Looks good!

tmcadam (approved):

PR #90, opened 2024-12-18
Create Tests for ESM1.6 Configs

### LINKED ISSUE #89: Create tests for `access-esm1.6-configs` repos
These tests should be about the same as the ESM1.5 ones, just with different constants.

CodeGat:
In this PR:
- **Add QA tests for ESM1.6 based on ESM1.5 - minus the DOI check as it doesn't yet exist**
- **Update general config test to have a `branch_type` fixture similar to the above QA tests, so we can skip tests that are on dev branches**
- **Add new pytest marker for ESM1.6**
- **Update package version to `0.0.8`**

Closes #89

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,7 +73,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [ "3.9", "3.10", "3.11" ]
+        python-version: [ "3.10", "3.11", "3.12" ]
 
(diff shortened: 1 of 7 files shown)


jo-basevi (approved):
LGTM. It'll be worth at some point adding the subclasses for `access-esm1.5` and `access-esm1.6` QA tests to avoid de-duplication of code.

PR #94, opened 2025-01-31
Sanitize Inputs for `!test`

### LINKED ISSUE #80: `config-comment-test`: Sanitize Newlines in Comment
## Background

See https://github.com/ACCESS-NRI/access-om3-configs/actions/runs/11431171056/job/31799782369#step:4:2

The workflow fails when we give a newline-separated comment command. We either need to sanitize the inputs or give a failure message. 

## Workaround

For now, keep comment commands on a single line.

CodeGat:
Closes #80

In this PR:

- **Take only the first line in a potentially multi-line comment as the command**
- **Use the commenter-permission-check action instead of defining it again**

### DIFF
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -131,7 +131,8 @@ jobs:
       - name: Get !test type
         id: test
         run: |
-          command="${{ github.event.comment.body }}"
+          # Only get the first line of the comment for processing as a command.
+          command=$(head -n 1 <<< "${{ github.event.comment.body }}")
           read -ra command_tokens <<< "$command"
           type_in_comment="${command_tokens[1]}"
           type_in_comment_valid=false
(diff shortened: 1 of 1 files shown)


atteggiani (approved):
Looks good! 

Only one very minor suggestion:
Instead of declaring the command to get the `command_tokens` multiple times:
```
# Only get the first line of the comment for processing as a command
command=$(head -n 1 <<< "${{ github.event.comment.body }}")
read -ra command_tokens <<< "$command"
```
A job could be set that outputs the command_tokens, so in the other jobs/steps you don't have to run/write those commands again but only use the output from the "get-command-tokens" job.

atteggiani:
Also, can't the logic:
command=$(head -n 1 <<< "${{ github.event.comment.body }}")
read -ra command_tokens <<< "$command"

CodeGat:
This is a good suggestion @atteggiani - I'll do that!

CodeGat:
Unfortunately, I've haven't been able to set the command tokens as a bash array either as an output or an environment variable, which means I don't actually save many lines...but I can simplify the command (as noted by your second comment)

PR #96, opened 2025-02-02
Delete and recreate `env.EXPERIMENTS_LOCATION` rather than wildcard

### LINKED ISSUE #95: Removing Repro Experiment Location Fails Due to Failed Wildcard Expansion
See https://github.com/ACCESS-NRI/access-esm1.5-configs/actions/runs/13064834305/job/36455296531?pr=122

This is due to the `rm` command being both double-quoted (to remove the possibility of an erroneous space in the command) and wildcarded (to delete everything in the directory). This means that the command is "remove a file called `*` from `${{ env.EXPERIMENT_LOCATION }}`. 

Update the `rm` logic to delete the directory itself, then recreate it.

CodeGat:
Closes #95

## Background

Rather than running into the issue of appropriately quoting and wildcarding a `rm`, delete the entire directory, then recreate it.

### DIFF
--- a/.github/workflows/test-repro.yml
+++ b/.github/workflows/test-repro.yml
@@ -111,7 +111,8 @@ jobs:
 
           # Remove base experiment (and everything else) if it exists
           if [ -d "${{ env.EXPERIMENT_LOCATION }}" ]; then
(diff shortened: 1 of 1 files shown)


blimlim (approved):
All looks good to me!

PR #98, opened 2025-02-05
Modify esm1p5/1p6 tests for AMIP configurations

### LINKED ISSUE #97: Qa tests for atmosphere only ESM1.5/1.6 configurations
ESM1.6 now has a new [development "AMIP" configuration](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-amip), and a similar configuration will soon be added for ESM1.5. These configurations use only the atmosphere model, reading in data files in place of the ocean and sea-ice models.

A few of the qa tests fail with these configurations - see [here for an example](https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/12900676632/job/35971561493). I think we'll need some small adjustments to the following tests:

1. [`test_config_start`](https://github.com/ACCESS-NRI/model-config-tests/blob/b536247d2123e4e00b6166370598b65830b446f6/src/model_config_tests/qa/test_access_esm1p6_config.py#L158):
   The AMIP configurations start on 1978-01-01. I think we could add a `branch.config_scenario == "amip"` option to this test with the 1978 start date.

2. [`test_cice_configuration_icefields_nml_in_ice_history_nml`](https://github.com/ACCESS-NRI/model-config-tests/blob/b536247d2123e4e00b6166370598b65830b446f6/src/model_config_tests/qa/test_access_esm1p6_config.py#L220C9-L220C65)

   This checks that the CICE namelists are correctly put together. We'll want to skip this test for the AMIP configurations as  there's no CICE sub-model.

We can leave the failing restart frequency tests as is. Payu can now do date based restart pruning with just the UM, and so for these I just need to update the configurations rather than the tests.

blimlim:
Closes #97 

This PR adds 1978-01-01 as a valid start date for the atmosphere-only (AMIP) configurations of ESM1.5 and 1.6. It also skips the `test_cice_configuration_icefields_nml_in_ice_history_nml` test for amip configurations, as these don't contain a CICE sub-model.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "model_config_tests"
-version = "0.0.9"
+version = "0.0.10"
 authors = [
   { name = "ACCESS-NRI" },
 ]
--- a/src/model_config_tests/qa/test_access_esm1p5_config.py
+++ b/src/model_config_tests/qa/test_access_esm1p5_config.py
@@ -28,6 +28,7 @@ VALID_NOMINAL_RESOLUTION: str = "100 km"
 VALID_REFERENCE: str = "https://doi.org/10.1071/ES19035"
 VALID_PREINDUSTRIAL_START: dict[str, int] = {"year": 101, "month": 1, "days": 1}
 VALID_HISTORICAL_START: dict[str, int] = {"year": 1850, "month": 1, "days": 1}
+VALID_AMIP_START: dict[str, int] = {"year": 1978, "month": 1, "days": 1}
 VALID_RUNTIME: dict[str, int] = {"years": 1, "months": 0, "days": 0}
 VALID_RESTART_FREQ: str = "10YS"
 VALID_MPPNCCOMBINE_EXE: str = "mppnccombine.spack"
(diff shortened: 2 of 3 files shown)


CodeGat (commented):
Looks good!

Only thing is, in the `pyproject.toml`, you'll have to increment the version before merging.

blimlim:
Thanks @CodeGat, would `0.0.10` be right for the new version?

CodeGat:
I think so... @jo-basevi, did we ever decide what the versions means in this repository? :laughing:

jo-basevi:
> I think so... @jo-basevi, did we ever decide what the versions means in this repository? 😆

No, I don't think we have.. I think the patch version increment makes sense, e.g. `0.0.10`, in this case. Though maybe for tests that will cause pre-existing configurations QA tests to fail, we should start incrementing the major and minor versions..

blimlim:
Sweet, just updated to `0.0.10` :)

CodeGat (approved):

CodeGat:
Looks good, give it a merge!

PR #101, opened 2025-02-11
Fix bug in ACCESS-OM3 reproducibility tests

### LINKED ISSUE #100: ACCESS-OM3 reproducibility test does not actually test reproducibility
The ACCESS-OM3 reproducibility tests currently use the contents of the output `ocean.stats` file to test reproducibility. However, by default the `ocean.stats` output is written daily. For the 3 hour runs, `ocean.stats` contains only stats about the initial state.

I'll follow up with a PR to ensure that the final state is also included in `ocean.stats`

See https://github.com/COSIMA/access-om3/issues/265

dougiesquire:
This PR ensure ocean.stats are written an the end of runs of less than one day. Closes #100

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "model_config_tests"
-version = "0.0.10"
+version = "0.0.11"
 authors = [
   { name = "ACCESS-NRI" },
 ]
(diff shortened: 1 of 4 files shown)


jo-basevi (approved):
Thanks, changes look good to me! The other option could be to override the default runtime for repro tests to a day (I think ACCESS-ESM1.5 does this) -  but if it supports 3 hours now then it makes sense to keep with that.

dougiesquire:
Thanks @jo-basevi. Yeah I think it makes sense to support short runs so the tests can run in a reasonable time. Our 0.25 deg config already takes a while, and high resolutions will be even slower.

Issue #99, opened 2025-02-07
More fine-grain control for QA checks on development branches


jo-basevi:
As brought up by @anton-seaice in https://github.com/ACCESS-NRI/access-om3-configs/issues/162, some of the QA tests are too onerous for configurations during development.

Currently with how the tests are set up, pytest markers are used to specify what tests to run. There is some configuration options in the `config/ci.json` file in the `main` branch of model-configs repositories (e.g., https://github.com/ACCESS-NRI/access-om3-configs/blob/main/config/ci.json and [some docs on how to modify this file](https://github.com/ACCESS-NRI/model-configs-template?tab=readme-ov-file#in-configcijson)). This file defines the default pytest markers to run for each style of tests (e.g. `qa`), and options to specify a specific marker for the target branch. E.g. the below config would use `"some_marker"` for QA pytests for PRs that are being merged into `dev-1deg_jra55do_ryf`:

```json
{
    # ...
    "qa": {
        "dev-1deg_jra55do_ryf": { 
            "markers": "some_marker"
        },
        "default": {
           "markers": "access_om3"
       }
}
```

One improvement would be to allow regex matching, e.g. "dev-*", rather than having to specify many branches. This will require some modifications to the action that parses this config files: https://github.com/ACCESS-NRI/model-config-tests/blob/main/.github/actions/parse-ci-config/action.yml

The other thing will be figuring out a list of QA checks that we might not want to run on dev branches - some examples are already listed in the above linked issue are:
- `test_runlog_is_on`
- `test_restart_freq_is_date_based`
- `test_metadata_contains_field` for 'reference' - maybe just have the metadata tests that check there isn't an experiment UUID?

I guess the most minimal change would be adding a `@pytest.mark.config_dev` (or similar) decorator above tests to run for configs early in development, and adding this marker here: 
https://github.com/ACCESS-NRI/model-config-tests/blob/8dc948e11f0935f2d764b699ac2e85c5bd4b1f8c/src/model_config_tests/conftest.py#L117-L119

anton-seaice:
> `@pytest.mark.config_dev` (or similar) decorator

I think we need this anyway, so that dev- and release- branches can use different markers. Would the marker be used for all `dev-` prefix branches in all configurations ? or is the meaning that these configurations are in alpha / beta development and the marker would be removed once that repository contains finalised released configurations?

[...]

PR #103, opened 2025-02-17
Changes to support OM3 development branches in QA checks

anton-seaice:
Contibrutes to #99 

This change splits the QA tests (the ones run on a github runner which don't run the model) into two sections. There is a section for Release branches and a section for Dev branches in configurations.

The dev branches tests are triggered via a new `dev_config` pytest marker. The behaviour of the existing `config` marker is unchanged, in that it runs both the tests for Release branches and those for Dev branches. The Dev branches test can be considered a subset of the tests applied for Release branches.

For consistency with other models , a test was added that the executable location is the spack release location.

@aidanheerdegen - mostly tagged you for a view on which tests should be dev vs release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 .idea
 /tests/tmp/
 coverage.xml
+/model-conf-test/
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@
 .idea
 /tests/tmp/
 coverage.xml
-/model-conf-test/
--- a/src/model_config_tests/conftest.py
+++ b/src/model_config_tests/conftest.py
@@ -115,7 +115,8 @@ def pytest_configure(config):
         "markers", "checksum_slow: mark tests as slow reproducibility tests"
     )
     config.addinivalue_line(
-        "markers", "config: mark as configuration tests for release branches in quick QA CI checks"
+        "markers",
+        "config: mark as configuration tests for release branches in quick QA CI checks",
     )
     config.addinivalue_line(
         "markers", "dev_config: mark as configuration tests in quick QA CI checks"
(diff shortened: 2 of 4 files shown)


aidanheerdegen:
Tests I think you definitely do want in `dev_config`:

`test_sync_is_not_enabled`
`test_sync_path_is_not_set`
`test_experiment_name_is_not_defined`
`test_metadata_does_contain_UUID` (should rename that to `test_metadata_does_not_contain_UUID` for clarity)


You don't want a shared experiment with a `sync` dir enabled. If you set the experiment name branching doesn't work and you definitely don't want the UUID defined in the `metadata.yaml` for a configuration. Not the case for an experiment though.

jo-basevi (commented):
Thanks for adding this in @anton-seaice!

Will the `access_om3` marker being used for the dev branches? If so, just double checking if that you are expecting `dev-*` branches to use a released model modules, e.g.:
```
modules:
    use:
        - /g/data/vk83/modules
    load:
        - access-om3/<version>
```
As the `test_access_om3_manifest_exe_in_release_spack_location` will check for a release Github artefact for a `spack.location` file. So this will fail for any `prerelease` modules.

anton-seaice:
Thanks @aidanheerdegen and @jo-basevi - i've addressed those comments

jo-basevi (dismissed):
Thanks, looks good to me!

aidanheerdegen (dismissed):
LGTM. Thanks.

dougiesquire (dismissed):
Not really needed at this point, but I did look and LGTM too

.gitignore
```diff
@@ -8,4 +8,5 @@
 .vscode
 .idea
 /tests/tmp/
-coverage.xml
\ No newline at end of file
+coverage.xml
+/model-conf-test/
```
anton-seaice, line 12:
```suggestion
```

anton-seaice:
Can someone approve again please ? 😂

jo-basevi (approved):

PR #108, opened 2025-02-19
Setup versioneer

### LINKED ISSUE #106: Add dynamic package versioning
Similarly to other `access-nri` python projects, it may be worth adding Versioneer to automate using git tags as package versions. 

Currently for deploying a new version of `model-config-tests` requires manually modifying the `version` in `pyproject.toml`, and then pushing a git tag that matches this version to the repository. Pushing a git tag will trigger the CD workflow that build sthe package and pushes it to PyPI.

jo-basevi:
Added [versioneer](https://github.com/python-versioneer/python-versioneer) support to allow dynamic versioning using git tags, rather than having to manually modify the version of the `pyproject.toml` file.

Closes #106


CodeGat (approved):
This will solve a lot of headaches :laughing: thank you!

PR #111, opened 2025-02-24
Support regex matching when parsing CI configuration file

jo-basevi:
Support regex matching for tag/branches when parsing `config/ci.json` file. This is to allow using regex `dev-.*` patterns to specify test configuration for all dev branches, e.g. 
```json
{
    # ...
    "qa": {
        "dev-.*": { 
            "markers": "dev_config"
        },
        "default": {
           "markers": "config or dev_config"
       }
   },
   "default": { 
        # ....
    }
}
```

I added a python script for parsing the file so I could just use the standard python regex library for string matching. This meant it was also easier to add tests.

The order for finding a key's (e.g. "markers") value now is: 
 1. Exact match for branch/tag name in test type (e.g. `qa`)
 2. Then first branch name matched with regex pattern for branch name in test type. This means a top-down ordering of regex patterns so would require more specific branch regex to be higher up in the list, e.g. `dev-1deg-.*` before `dev-.*`
 3. Default for the test type (e.g. 'qa') 
 4. Top-level default

To avoid the composite action modifying the caller workflow environment, I've set `update-environment: false` in `setup-python` action, and called the python executable directly using the output of the action (see https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-update-environment-flag).

Similarly to avoid checking out the `model-config-tests` repository to obtain the `parse-file.py` file of the correct version, I am using `github.action_path` to access the files in the `.github/actions/pares-ci-config` directory (see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context).

References #99

### DIFF
--- a/.github/actions/parse-ci-config/parse_file.py
+++ b/.github/actions/parse-ci-config/parse_file.py
@@ -1,0 +1,113 @@
+import argparse
+import json
+import re
+
+
+def get_config_value(config, test_type, reference, key):
+    """
+    Retrieve the value for a given key from the nested structure of the JSON file.
+    It first checks the specific 'test_type' and 'reference', then falls back to the default values if not found.
+
+    Parameters:
+    config (dict): The dictionary containing the CI configuration.
+    test_type (str): Type of check/test to run (e.g., "reproducibility", "qa", or "scheduled").
+    reference (str): Name of Git branch or tag to run CI testing on.
+    key (str): The key to retrieve the value for (e.g. "python-version").
+
+    Returns:
+    str: The value associated with the key.
+    """
+    # Check for exact match
+    value = config.get(test_type, {}).get(reference, {}).get(key)
+    if value:
+        return value
+
+    # Check for branch/tag with the longest regex match
+    # E.g. dev-branch-1 should match dev-branch-* over dev-*
+    longest_match = None
+    longest_match_length = 0
+    for pattern in config.get(test_type, {}):
+        match = re.match(pattern, reference)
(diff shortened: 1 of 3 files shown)


.github/actions/parse-ci-config/parse_file.py
```diff
@@ -0,0 +1,108 @@
...
+    Returns:
+    str: The value associated with the key.
+    """
+    # Check for exact match of test_type and reference
+    value = config.get(test_type, {}).get(reference, {}).get(key)
+    if value:
+        return value
+
+    # Check for reference regex match with a key defined with priority given
+    # to the first match found (so top-down order matters)
```
CodeGat, line 28: I was a bit worried about this bit - Order isn't guaranteed to be maintained in a JSON object usually. 
Luckily the default when using [`json.dump`](https://docs.python.org/3/library/json.html#json.dump) is `sort_keys=False`, which means that object order is maintained when being parsed into python dictionaries. Phew! Might be something worth keeping in mind if it's going to be de/serialized outside of python one day...

CodeGat (approved):

.github/actions/parse-ci-config/parse_file.py
jo-basevi: Yeah I wasn't too sure on it initially too, so had to run a couple of tests to double check that order was being preserved when reading json files.

I wasn't really sure on how else to prioritise one regex over another either..

PR #113, opened 2025-02-24
Support branch version for model-config-tests in test configuration

### LINKED ISSUE #109: Support using main branch of model-config-tests in configs
Thanks @anton-seaice for the suggestion of adding a `main` option for `model-config-tests-version` in the CI configuration file `config/ci.json`. This would allow the option to have the latest tests run for particular branches - e.g. `dev` branches (see https://github.com/ACCESS-NRI/model-config-tests/issues/99 for allowing regex for `dev-*` branches in `config/ci.json`).

This will require logic when each time `model-config-tests` is installed, e.g. `pip install model-config-tests==<version>` to instead install the main branch of `model-config-tests` from github - `pip install git+https://github.com/ACCESS-NRI/model-config-tests.git@main`

I think using a fixed version of `model-config-tests` should be the default or recommended in general, so to allow for any breaking changes in `model-config-tests` updates.

jo-basevi:
To support setting `"model-config-tests-version": "main"` in `config/ci.json`, this PR adds some logic to attempt to install `model-config-tests` first from PyPI. Then if that fails, it attempts to install `model-config-tests` directly from Github, using the version as the reference. This means it will support git commits and other branch names if it's ever needed for testing.

Closes #109

### DIFF
--- a/.github/workflows/config-pr-1-ci.yml
+++ b/.github/workflows/config-pr-1-ci.yml
@@ -184,7 +184,13 @@ jobs:
           cache-dependency-path: model-config-tests/pyproject.toml
 
       - name: Install model-config-tests
-        run: pip install model-config-tests=='${{ needs.config.outputs.qa-model-config-tests-version }}'
+        run: |
+          VERSION="${{ needs.config.outputs.qa-model-config-tests-version }}"
+          pip install "model-config-tests==${VERSION}" || {
+            echo "Failed to install model-config-tests==${VERSION} from PyPI"
+            echo "Attempting to install directly from Github repository with reference ${VERSION}"
+            pip install git+https://github.com/access-nri/model-config-tests.git@${VERSION}
+          }
 
       - name: Invoke Simple CI Pytests
         # We continue on error because we will let the checks generated in
--- a/.github/workflows/test-repro.yml
+++ b/.github/workflows/test-repro.yml
@@ -153,7 +153,12 @@ jobs:
           source "${{ env.TEST_VENV_LOCATION }}/bin/activate"
 
           # Install model-config-tests
-          pip install model-config-tests==${{ inputs.model-config-tests-version }}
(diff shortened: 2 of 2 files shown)


CodeGat (approved):
Looks good to me! 

In what cases would be want to be running an unreleased version of `model-config-tests`? Is it for temporary testing of the `model-config-tests` package in the model configs repositories that would use it?

CodeGat:
Cool! Looks good to me then

.github/workflows/config-pr-1-ci.yml
```diff
@@ -184,7 +184,13 @@ jobs:
...
 
       - name: Install model-config-tests
-        run: pip install model-config-tests=='${{ needs.config.outputs.qa-model-config-tests-version }}'
+        run: |
+          VERSION="${{ needs.config.outputs.qa-model-config-tests-version }}"
+          pip install "model-config-tests==${VERSION}" || {
+            echo "Failed to install model-config-tests==${VERSION} from PyPI"
+            echo "Attempting to install directly from Github repository with reference ${VERSION}"
+            pip install git+https://github.com/access-nri/model-config-tests.git@${VERSION}
+          }
```
aidanheerdegen, line 193: Just checking, if the `pip install` fails the CI will also fail right?

jo-basevi: Yeah if the pip install from Github also fails, then the CI will fail (tested with running on a github runner). I assume it would be similar on the ssh into NCI step in `test-repro` but I might just double check that now..

jo-basevi: Ok worth the check as my assumptions were wrong with the ssh section in `test-repro.yml`.. I've opened a PR to exit immediately if there are errors https://github.com/ACCESS-NRI/model-config-tests/pull/116

PR #116, opened 2025-02-25
test-repro.yml: Exit ssh session if test setup fails

jo-basevi:
During the ssh session to gadi which runs the repro tests, if any of the test setup fails, e.g. loading `payu` modules or pip installing `model-config-tests`, it should exit immediately and report as a failed step. At the moment, I've just added a `set -e` at the start of the script which seems to work when testing on a separate test repository.


aidanheerdegen (approved):
LGTM

PR #120, opened 2025-02-27
Skip config test checking UUID in `metadata.yaml` if `metadata.yaml`doesn't exist

### LINKED ISSUE #119: `qa/test_config.py`: only check `metadata.yaml` with `config` marker
Currently, the config test to check that the `metadata.yaml` does not contain a UUID is run with both the `config` and `dev_config` markers:

https://github.com/ACCESS-NRI/model-config-tests/blob/eed3efffa198c03f52f3ddf8089493f6b2e9a3f6/src/model_config_tests/qa/test_config.py#L215-L219

However, many dev configs don't contain a `metadata.yaml`, causing this test to fail - e.g. see https://github.com/ACCESS-NRI/access-om3-configs/pull/197. We should ~move the `test_metadata_does_not_contain_UUID` test into the `TestRelConfig` class so that it is only run with the `config` marker~ only run the `test_metadata_does_not_contain_UUID` on dev configs if the `metadata.yaml` exists.

I'll follow up with a PR.

dougiesquire:
Closes #119

### DIFF
--- a/src/model_config_tests/conftest.py
+++ b/src/model_config_tests/conftest.py
@@ -50,6 +50,13 @@ def metadata(control_path: Path):
 
 
 @pytest.fixture(scope="session")
+def skipif_no_metadata(control_path):
+    metadata_path = control_path / "metadata.yaml"
+    if not metadata_path.exists():
+        pytest.skip("No metadata.yaml file exists")
+
+
+@pytest.fixture(scope="session")
 def config(control_path: Path):
     """Read the config file in the control directory"""
     config_path = control_path / "config.yaml"
--- a/src/model_config_tests/qa/test_config.py
+++ b/src/model_config_tests/qa/test_config.py
@@ -212,10 +212,11 @@ class TestConfig:
                 "The target branch is a dev version and doesn't require a stable module location"
             )
 
+    @pytest.mark.usefixtures("skipif_no_metadata")
     def test_metadata_does_not_contain_UUID(self, metadata):
         assert "experiment_uuid" not in metadata, (
             "`experiment_uuid` should not be defined in metadata, "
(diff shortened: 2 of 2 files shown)


anton-seaice (changes requested):
Sorry - we intentionally have this test in the `dev_config` marker so that folks don't commit these into configs.

It needs a special case for handling if `metadata.yaml` does not exist.

dougiesquire:
Good point - rushed this without really thinking

dougiesquire:
This was actually a bit fiddly because the `metadata.yaml` is loaded in a fixture. I've added a fixture to skip the test if there is no `metadata.yaml` in the control dir

jo-basevi (approved):
I didn't know you could apply a fixture `@pytest.mark.usefixtures("skipif_no_metadata")` before the fixture in the function call is run so that's good to learn. It looks good to me

aidanheerdegen:
Are we insisting that there be a `metadata.yaml` file for release configurations? Because I definitely think we should.

dougiesquire:
> I didn't know you could apply a fixture @pytest.mark.usefixtures("skipif_no_metadata") before the fixture in the function call is run so that's good to learn. It looks good to me

Neither did I!

> Are we insisting that there be a metadata.yaml file for release configurations? Because I definitely think we should.

Yup, there are other tests run with the `config` marker that insist there is a `metadata.yaml`

dougiesquire:
@anton-seaice this requires approval from you in order to merge

anton-seaice:
Thanks @dougiesquire 

Did adhoc tests without metadata.yaml - works, and with metadata.yaml - still fails correctly if there is a UUID defined :-)

anton-seaice (approved):

PR #134, opened 2025-03-18
ESM1.6 qa checks for io_layout

### LINKED ISSUE #133: Allow collation to be disabled for ESM1.6 qa tests
Collation is now disabled in the esm1.6 development configuration: https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/55

The `test_mppnccombine_fast_collate_exe` should be updated to allow for collation to be disabled.

https://github.com/ACCESS-NRI/model-config-tests/blob/5c12d7c2341e7737d8af4ef07120753ca9061ca6/src/model_config_tests/qa/test_access_esm1p6_config.py#L206-L223

blimlim:
Closes #133.

This PR updates the QA tests for ESM1.6. Previously, the `test_mppnccombine_fast_collate_exe` test checked that the collation executable and mpi setting were correctly set.

https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/55 removed collation for the configuration, and set `io_layout=1,1` in the ocean `input.nml` file. These two changes are linked, and are required to go together (i.e. switching off collation only makes sense if you have `io_layout=1,1`. Because of this, I've modified the tests to check that collation is disabled and correct `io_layout` is set.

Any feedback or suggestions are very welcome!

### DIFF
--- a/src/model_config_tests/qa/test_access_esm1p6_config.py
+++ b/src/model_config_tests/qa/test_access_esm1p6_config.py
@@ -214,9 +214,7 @@ class TestAccessEsm1p6:
         if branch.config_scenario == "amip":
             pytest.skip("amip scenarios do not contain the MOM sub-model")
 
-        assert "collate" in config, error_field_nonexistence(
-            "collate", "config.yaml"
-        )
+        assert "collate" in config, error_field_nonexistence("collate", "config.yaml")
 
         assert not config["collate"]["enable"], error_field_incorrect(
             "collate.enable", "config.yaml", False
@@ -246,15 +244,13 @@ class TestAccessEsm1p6:
 
         mom_input = f90nml.read(mom_input_path)
 
-        assert "io_layout" in mom_input[OCEAN_MODEL_NML_NAME], (
-            error_field_nonexistence("io_layout", MOM_INPUT_NML_FNAME)
+        assert "io_layout" in mom_input[OCEAN_MODEL_NML_NAME], error_field_nonexistence(
+            "io_layout", MOM_INPUT_NML_FNAME
         )
-        assert mom_input["ocean_model_nml"]["io_layout"] == VALID_IO_LAYOUT, (
-            error_field_incorrect(
-                "io_layout",
-                MOM_INPUT_NML_FNAME,
-                ','.join(i for i in VALID_IO_LAYOUT)
-            )
+        assert (
+            mom_input["ocean_model_nml"]["io_layout"] == VALID_IO_LAYOUT
+        ), error_field_incorrect(
+            "io_layout", MOM_INPUT_NML_FNAME, ",".join(i for i in VALID_IO_LAYOUT)
         )
 
     def test_cice_configuration_icefields_nml_in_ice_history_nml(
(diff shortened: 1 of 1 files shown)


src/model_config_tests/qa/test_access_esm1p6_config.py
```diff
@@ -203,24 +207,51 @@ def test_config_restart_freq(self, config):
...
-            )
+        assert "collate" in config, error_field_nonexistence("collate", "config.yaml")
 
-            # Loading the yaml into a dict also converts
-            # `collate.mpi:true`/`collate.mpi:false` to `True`/`False` so we
-            # can assert if it is a `bool`.
-            assert isinstance(config["collate"]["mpi"], bool), error_field_incorrect(
-                "collate.mpi", "config.yaml", "true or false"
-            )
+        assert not config["collate"]["enable"], error_field_incorrect(
```
dougiesquire, line 219: Does `enable` have to be explicitly set for collation to be enabled?

I think this would still run collation, e.g.:
```
collate:
    exe: mppnccombine.spack
    restart: true
    mem: 4GB
    walltime: 1:30:00
    mpi: false
```

Also, I think the lack of a `"collate"` key would also mean no collation?

src/model_config_tests/qa/test_access_esm1p6_config.py
```diff
@@ -203,24 +207,51 @@ def test_config_restart_freq(self, config):
...
+
+        mom_input = f90nml.read(mom_input_path)
+
+        assert "io_layout" in mom_input[OCEAN_MODEL_NML_NAME], error_field_nonexistence(
+            "io_layout", MOM_INPUT_NML_FNAME
+        )
+        assert (
+            mom_input["ocean_model_nml"]["io_layout"] == VALID_IO_LAYOUT
+        ), error_field_incorrect(
+            "io_layout", MOM_INPUT_NML_FNAME, ",".join(i for i in VALID_IO_LAYOUT)
```
dougiesquire, line 253:
```suggestion
            "io_layout", MOM_INPUT_NML_FNAME, ",".join(str(i) for i in VALID_IO_LAYOUT)
```

dougiesquire (changes requested):
Thanks @blimlim! Sorry for taking so long - got sucked into a meeting

src/model_config_tests/qa/test_access_esm1p6_config.py
blimlim: Unfortunately payu's default behaviour is to enable collation when there's no `collate` keyword

https://github.com/payu-org/payu/blob/39e4b9bdabceca3836bc25f0fcfe411d0c0b9f13/payu/experiment.py#L839

blimlim:
Thanks @dougiesquire and not a problem at all! Good point with the `str(i)` change. From what I can tell, I think I need to keep the explicit check that `["collation"]["enable"]` is `False`. I could add an extra check to make sure the `[enable]` flag is present too. Let me know what you think!

src/model_config_tests/qa/test_access_esm1p6_config.py
dougiesquire: Snap!

blimlim: Yeah can do!

blimlim: Sweet, added in and ready for review!

dougiesquire (approved):
Thanks @blimlim!

PR #135, opened 2025-03-18
Add Gadi.spack.location download url to QA executable test

### LINKED ISSUE #132: Fix `spack_location` test to support `build-cd`'s v4
As reported in https://github.com/ACCESS-NRI/access-om3-configs/pull/220#issuecomment-2719573584, the test that checks whether the executables in the  manifests match an install path in the `spack.location` release artefact - is using the wrong Github url. In `build-cd`'s `v4` adds in support for deployment to multiple different HPC systems, and I think the filenames now include the deployment environment, e.g. `Gadi`.

I think a quick fix would be to check if either of the releases exist, e.g
https://github.com/ACCESS-NRI/model-config-tests/blob/5c12d7c2341e7737d8af4ef07120753ca9061ca6/src/model_config_tests/qa/test_config.py#L304-L307

or

```python
url = ( 
     f"https://github.com/ACCESS-NRI/{model_repo_name}/" 
     f"releases/download/{module_version}/Gadi.spack.location" 
 ) 
```

jo-basevi:
This PR attempts to download a `Gadi.spack.location` file in the release artefact, if a `spack.location` doesn't exist. This is to support model releases deployed to Gadi using `ACCESS-NRI/build-cd` v4. 

I think it'll need a rethink once configurations are developed to run on other HPCs

Closes #132

### DIFF
--- a/src/model_config_tests/qa/test_config.py
+++ b/src/model_config_tests/qa/test_config.py
@@ -259,6 +259,40 @@ def read_config_model_exes(config: dict[str, Any]):
     return exes
 
 
+def get_spack_location_file(model_repo_name, model_version):
+    """Return the spack.location file for the model version
+    from a Github release artefact. Raises an AssertionError if the
+    release artefact or spack.location file is not found."""
+    base_url = f"https://github.com/ACCESS-NRI/{model_repo_name}/releases"
+    # Check whether there is a release artefact for the model version
+    release_url = f"{base_url}/tag/{model_version}"
+    assert (
+        requests.get(release_url).status_code == 200
+    ), f"Failed to find release artefact for model version at {release_url}"
+
+    # Urls for spack.location file in release artefacts assets,
+    # Note: Gadi.spack.location filename is used for models built with
+    # access-nri/build-cd version v4 and later
+    urls = [
+        f"{base_url}/download/{model_version}/spack.location",
+        f"{base_url}/download/{model_version}/Gadi.spack.location",
+    ]
+
+    # Attempt to download a spack.location file
+    spack_location = None
+    for url in urls:
+        response = requests.get(url)
+        if response.status_code == 200:
+            spack_location = str(response.content)
+
+    assert spack_location is not None, (
+        "Failed to download a spack.location or Gadi.spack.location file in "
+        f"the release artefact for model version {model_version}. "
+        f"Checked urls: {(', ').join(urls)}"
+    )
+    return spack_location
+
+
 def check_manifest_exes_in_spack_location(
     model_module_name, model_repo_name, control_path, config
 ):
@@ -301,15 +335,7 @@ def check_manifest_exes_in_spack_location(
     _, module_version = modules[0].split("/")
 
     # Use the module version to download spack.location file
-    url = (
-        f"https://github.com/ACCESS-NRI/{model_repo_name}/"
-        f"releases/download/{module_version}/spack.location"
-    )
-
-    # Check there is a released spack.location file for the module version
-    response = requests.get(url)
-    assert response.status_code == 200, f"Failed to download spack.location from {url}"
-    spack_location = response.content
+    spack_location = get_spack_location_file(model_repo_name, module_version)
 
     # Read exe full paths in the manifests
     exe_paths = read_exe_manifest_fullpaths(control_path)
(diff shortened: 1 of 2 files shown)


src/model_config_tests/qa/test_config.py
```diff
@@ -259,6 +259,40 @@ def read_config_model_exes(config: dict[str, Any]):
...
+    for url in urls:
+        response = requests.get(url)
+        if response.status_code == 200:
+            spack_location = str(response.content)
+
+    assert spack_location is not None, (
+        "Failed to download a spack.location or Gadi.spack.location file in "
+        f"the release artefact for model version {model_version}. "
+        f"Checked urls: {(', ').join(urls)}"
+    )
```
atteggiani, line 292: The test for this portion is not present.
Is it not required?

atteggiani (dismissed):
Looks good.

Just a minor comment about missing test for a portion of code, but not necessarily relevant.

src/model_config_tests/qa/test_config.py
jo-basevi: I've added in a test with some mocking so that code can be reached in [df4732b](https://github.com/ACCESS-NRI/model-config-tests/pull/135/commits/df4732bb041a0d86d9443a2c568651f65a10cff7)

tests/qa/test_test_config.py
```diff
@@ -29,3 +37,53 @@ def test_test_config_access_om2():
...
+    return response
+
+
+def test_get_spack_location_file_no_spack_location():
+    """
+    Test that an error is raised when the spack.location
+    or Gadi.spack.location file is not found in the release artefact
+    """
+    with patch("requests.get", side_effect=mock_request_get):
+        error_msg = r"Failed to download a spack.location .*"
```
atteggiani, line 87:
```suggestion
        error_msg = r"Failed to download a spack\.location .*"
```
Very minor

tests/qa/test_test_config.py
```diff
@@ -29,3 +37,53 @@ def test_test_config_access_om2():
...
+def mock_request_get(url, *args, **kwargs):
+    """Custom side effect function for mocking requests.get to pass the initial
+    initial request (e.g that a release artefact exists)
+    but returns a 404 status for subsequent requests.get calls
+    """
+    response = Mock()
+    if url == "https://github.com/ACCESS-NRI/fake-repo/releases/tag/fake-version":
+        response.status_code = 200
+    else:
+        response.status_code = 404
```
atteggiani, line 77:
```suggestion
    if url == "https://github.com/ACCESS-NRI/fake-repo/releases/tag/fake-version":
        response.status_code = 200
    else:
        response.status_code = 404
```
Why do you need this differentiation within this test? 
Are you planning to use `mock_request_get` in other tests?

Otherwise I would simply set `response.status_code = 404` (or any number but `200`)

atteggiani: Oh I see now you're using `requests` also above.

jo-basevi: Yeah, I use requests to check if a release artefact exists first before trying to download a file

src/model_config_tests/qa/test_config.py
```diff
@@ -259,6 +259,40 @@ def read_config_model_exes(config: dict[str, Any]):
...
+        f"{base_url}/download/{model_version}/spack.location",
+        f"{base_url}/download/{model_version}/Gadi.spack.location",
+    ]
+
+    # Attempt to download a spack.location file
+    spack_location = None
+    for url in urls:
+        response = requests.get(url)
+        if response.status_code == 200:
+            spack_location = str(response.content)
```
atteggiani, line 286: I also just noticed, can both `urls` return a status_code == 200? 
Otherwise you might break the for loop and exit soon after the first loop. 
Again very minor.

atteggiani (approved):

src/model_config_tests/qa/test_config.py
jo-basevi: I don't think it can.. So earlier versions of build-cd (< `v4`) just release a `spack.location` and later versions, have a format of `HPC.spack.location`. If there somehow was both those files in the release, I think `Gadi.spack.location` would be the preferred one?

atteggiani: Well, I am not sure about that, but if that's the case the current code works fine. 

If only one among the urls could give a good response at the same time, my solution would be slightly faster (at times) but it's a very negligible factor. 

Therefore, I think you can safely keep it as it is.

jo-basevi: I doubled checked with Tommy if there could ever be both `spack.location` and a `Gadi.spack.location` and he reckons there wouldn't be, and it would potentially cause errors at the model build CD stage.

jo-basevi:
Thanks for the review @atteggiani!

PR #137, opened 2025-03-18
Add test-specific repro test markers

### LINKED ISSUE #136: Additional repro test markers
The reproducibility tests currently have two markers:
- `checksum`, which runs the historical repro check
- `checksum_slow`, which runs the repeat and restart repro checks

In some cases it would be helpful to have a finer level of control over which tests are run. Could we add some additional markers to allow users to request specific tests?

dougiesquire:
This PR adds some test-specific markers to the reproducibility tests to allow finer-grain control over which repro tests are run.

Closes #136

### DIFF
--- a/src/model_config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/test_bit_reproducibility.py
@@ -47,7 +47,7 @@ def read_historical_checksums(
 
 class TestBitReproducibility:
 
-    @pytest.mark.checksum
+    @pytest.mark.repro
     @pytest.mark.repro_historical
     def test_bit_repro_historical(
         self,
@@ -133,8 +133,9 @@ class TestBitReproducibility:
             hist_checksums == checksums
         ), f"Checksums were not equal. The new checksums have been written to {checksum_output_file}."
 
-    @pytest.mark.checksum_slow
+    @pytest.mark.repro
     @pytest.mark.repro_repeat
+    @pytest.mark.slow
     def test_bit_repro_repeat(self, output_path: Path, control_path: Path):
         """
         Test that a run has same checksums when ran twice
@@ -156,8 +157,9 @@ class TestBitReproducibility:
 
         assert produced == expected
 
-    @pytest.mark.checksum_slow
+    @pytest.mark.repro
     @pytest.mark.repro_restart
+    @pytest.mark.slow
     def test_restart_repro(self, output_path: Path, control_path: Path):
         """
         Test that a run reproduces across restarts.
--- a/src/model_config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/test_bit_reproducibility.py
@@ -48,6 +48,7 @@ def read_historical_checksums(
 class TestBitReproducibility:
 
     @pytest.mark.checksum
+    @pytest.mark.repro_historical
     def test_bit_repro_historical(
         self,
         output_path: Path,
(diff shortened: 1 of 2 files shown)


src/model_config_tests/test_bit_reproducibility.py
```diff
@@ -133,6 +134,7 @@ def test_bit_repro_historical(
         ), f"Checksums were not equal. The new checksums have been written to {checksum_output_file}."
 
     @pytest.mark.checksum_slow
+    @pytest.mark.repro_repeat
```
aidanheerdegen, line 137: We're starting to get a lot of quite specific markers.

I wonder if we wouldn't be better having separate markers, e.g.
```python
@pytest.mark.checksum
@pytest.mark.repro
@pytest.mark.historical
```
and
```python
@pytest.mark.checksum
@pytest.mark.repro
@pytest.mark.slow
@pytest.mark.repeat
```

Then use select logic to get very specific tests,  e.g. `checksum and repro and historical`, or more general, e.g. `checksum and slow`.

Note I have never studied formal logic, so am happy to be schooled why this is a terrible idea.

dougiesquire: I'm not sure I understand the distinction between the `@pytest.mark.checksum` and `@pytest.mark.repro` markers?

Has this reduced the number of markers or increased it?

dougiesquire: Other than some marker name changes, does the difference between what I've done and what you're suggesting boil down to splitting:

```
@pytest.mark.checksum_slow
```
into
```
@pytest.mark.checksum
@pytest.mark.slow
```
?

I agree that seems like a good idea.

jo-basevi: The above looks good to me. I am for using`@pytest.mark.repro` for each test (rather than `@pytest.mark.checksum`) as it's a bit more descriptive.

dougiesquire: Thanks @jo-basevi, I agree re `@pytest.mark.repro`. Do you have a preference for, for example, `repro_repeat` vs `repeat`? The former feels redundant but the latter lacks clarity.

jo-basevi: I feel `repro_repeat` has less risk of marker collision (though I doubt there will be `repeat` markers anyway for QA tests..). Also if the marker is used directly on it's own, `repro_repeat` makes more sense to me. Though I don't really feel strongly either way.

dougiesquire: Addressed in [0e96a1e](https://github.com/ACCESS-NRI/model-config-tests/pull/137/commits/0e96a1e145a6a67212869c6e74ae3da80deb7836)

jo-basevi (approved):
Thanks @dougiesquire, looks good to me

dougiesquire:
Because we've now renamed markers, I'll need to update config repos that are configured to use `model-config-tests` `main` as soon as this is merged.

It's probably also worth tagging another release, so we don't end up with a horrible mix of markers in the `ci.json` - e.g. https://github.com/ACCESS-NRI/access-om3-configs/blob/main/config/ci.json

jo-basevi:
Good point, I agree that a new release would be good - it's probably time to actually start incrementing the major version now that old markers wouldn't work..

dougiesquire:
Okay so I'll tag v0.1.0 once this is merged? I'm going to get the PRs to the config repos ready to go to minimize the period that things don't work.

aidanheerdegen (approved):
LGTM2

PR #139, opened 2025-03-19
`!test repro commit`: Only commit if checksums differ to the ones on the branch

### LINKED ISSUE #138: `!test repro commit` can't commit when the checksum in the repo is identical
See https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2734966108

## Background

The `!test repro commit` command tests the current PR `HEAD` against the commit in the base branch (not the commit before).
The repro check fails, as the checksums generated in the PR `HEAD` are different from the base branch.
We attempt to commmit these changes, but because of an earlier commit on the branch (which was a previous invocation of !test repro commit which committed the same checksum successfully), we have no changes to add and therefore the git push fails.

Since the `!test repro commit` only tests the current HEAD against the base branch, rather than the current HEAD against a previous commit, the commit fails due to identical checksums stored on the PR branch.

## Implementation Plan

* The check against the base branch commit rather than the previous commit is expected behaviour
* The `git commit` can be handled more gracefully if there are identical checksums stored on the branch.

CodeGat:
Closes #138

## Background

If an earlier commit in a PR branch updates the checksums, and a later `!test repro commit` command finds that the base branch checksums differ, it will attempt to commit the new checksums. If these new checksums are actually identical to an earlier invocation of `!test repro commit` on the branch, it will fail to commit and push the changes. 

This PR checks that changes are actually added before committing and pushing. 

## This PR

In this PR:

* Only commit and push if there are actually checksums to add on this branch
* Update the repro result comment

## Testing

Tested the `git status --porcelain` (and indeed the whole step) locally to verify it is what we want. The `--porcelain` bit essentially makes the output machine-readable and never changes between versions of `git`, making it useful for scripts.

### DIFF
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -313,7 +313,7 @@ jobs:
 
           if [ -z "$(git status --porcelain)" ]; then  # If there are no added files
             echo "::warning::Attempting to add the changed checksums failed as they were identical to the existing ones on the PR branch. A previous commit must have updated them already, and they match."
-            continue
+            exit 0
           fi
 
           git commit -m "Updated checksums as part of ${{ env.RUN_URL }}"
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -246,7 +246,7 @@ jobs:
             - `${{ needs.prepare-command.outputs.config-ref }}` (checksums created using commit ${{ env.CONFIG_REF_URL }}), against
             - `${{ needs.prepare-command.outputs.compared-config-ref }}` (checksums in commit ${{ env.COMPARED_CONFIG_REF_URL }})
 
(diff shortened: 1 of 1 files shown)


.github/workflows/config-comment-test.yml
```diff
@@ -310,6 +310,12 @@ jobs:
       - name: Commit and Push Updates
         run: |
           git add .
+
+          if [ -z "$(git status --porcelain)" ]; then  # If there are no added files
+            echo "::warning::Attempting to add the changed checksums failed as they were identical to the existing ones on the PR branch. A previous commit must have updated them already, and they match."
+            continue
```
atteggiani, line 316: I don't think `continue` here is needed.
If the behaviour should be to exit with a success, I would change it to `exit 0` instead.

atteggiani (commented):
Overall looks good. 
I just added a comment.

.github/workflows/config-comment-test.yml
CodeGat: Updated, thank you @atteggiani

atteggiani (approved):
LGTM

PR #142, opened 2025-03-20
Run QA checks for bot commits

### LINKED ISSUE #141: For `access-bot` repro result commits, run the QA tests
Currently, when `access-bot` commits repro changes, it doesn't do the QA checks. If the QA checks fail in an earlier commit, this result is not maintained by the next bot commit, which could confuse users, as the check rollup shows a tick where the previous one failed. This is because the QA test is skipped, rather than passing, but confusing nonetheless. 

See https://github.com/ACCESS-NRI/access-om3-configs/pull/220/commits/ce75d880af94d249bf6a0b0e28daf8f43a0d2941 vs https://github.com/ACCESS-NRI/access-om3-configs/pull/220/commits/08c8b87581e0028b46eb0118d5280106a19ad494

CodeGat:
Closes #141

## Background

Bot commits don't do QA checks, while regular commits do. This might confuse users because they will get a passing check rollup where they expected a failing one, because the QA checks fail in an earlier commit, and "pass" (i.e., are skipped) in a later bot one. 

## In this PR

* Remove the `commit-check` dependent job for `qa-ci`

### DIFF
--- a/.github/workflows/config-pr-1-ci.yml
+++ b/.github/workflows/config-pr-1-ci.yml
@@ -153,9 +153,7 @@ jobs:
     # Run quick, non-HPC tests on the runner.
     name: QA CI Checks
     needs:
-      - commit-check
       - config
-    if: needs.commit-check.outputs.authorship != vars.GH_ACTIONS_BOT_GIT_USER_NAME
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
Looks good to me

PR #144, opened 2025-03-24
Add reproducibility tests for ESM1.5/1.6 AMIP configurations

### LINKED ISSUE #143: Reproducibility tests for ESM1.5/1.6 AMIP configurations
The [amip configuration](https://github.com/ACCESS-NRI/access-esm1.5-configs/tree/amip) for ESM1.5 is ready to be turned into a development branch before being released. The [dev-amip](https://github.com/ACCESS-NRI/access-esm1.6-configs/tree/dev-amip) branch is also being used for testing and development of ESM1.6.

Both of these configurations need reproducibility tests enabled, however new tests will need to be created. The AMIP configurations only don't contain the mom submodel, and hence do not produce checksums of ocean variables at the end of the simulation.

The atmosphere logs do contain a `Final Absolute Norm` that's written at each timestep. This is a numerical solver statistic, and @MartinDix reports that changes in the atmospheric state will cause changes in these final absolute norms. These could be extracted and used in the existing checksum framework.

blimlim:
Closes #143.

This pull request enables repro testing for the ESM1.5/1.6 AMIP configurations. These configurations do not contain the mom5 submodel, and hence separate information from the um7 logs must be used to compare simulation results.

This PR adds UM7 specific code, which extracts the last 10 `Final Absolute Norms` from the atmosphere log file, and writes them to a checksum file of the usual format. The UM writes this `Final Absolute Norm` for every timestep. It's a statistic from the numerical solver, and is highly sensitive to the atmospheric state.  It's technically not a "checksum", but can play a very similar role.

The choice to use the last 10 is a bit arbitrary. I thought it best to save more than one, to reduce the risk of collisions, while saving the value for every single timestep seemed excessive.

The other changes made by this PR are during the experiment setup and during the checksum extraction, where parts of the code are now only run for configurations containing the relevant submodels.

Would be great to hear any feedback or suggestions!


blimlim:
Looks like I need to update the unit tests. @jo-basevi, will let you know when I've done that!

blimlim:
@jo-basevi, I've had a go at getting the tests working, though am not 100% happy with what I came up with. 

The issue stemmed from these changes to the `extract_checksums` method for the `AccessEsm1p5` model
```diff 
+        submodels = {
+           submodel["model"]: {"submodel_name": submodel["name"]}
+           for submodel in self.experiment.config["submodels"]
+        }
```
The AMIP configurations have fewer submodels in their `config.yaml` files, and extract checksums from a filepath determined by the `submodel["name"]` value in the `config.yaml`. I found it necessary to read in the configuration data to get `extract_checksums` working with both the amip and non-amip configurations.

On the unit-test side, this meant providing configuration information to `test_extract_checksums` where necessary. This makes that test a bit more complicated than before, and still leaves a bit of a gap: I don't know if it's currently possible to test the checksum extraction methods for both a non-amip and an amip configuration for the same model, as the resources directory has a single output and checksum directory for each model.

A couple of options I could think of: 
1. The amip configurations are a weird case, other models won't run into similar issues. I could write a separate test for them completely, though this is a little messy.
2. Set up per-configuration output and checksum resources. This might unnecessarily complicate the structure, especially since for the other models, the configurations are irrelevant for the unit tests.
3. Rather than using the same payu model class for the amip configurations, create new model drivers (e.g. `access-esm1.6-amip`), though I think this could be a bit too much and worry that it could cause other complications down the line.

Would be keen to hear any ideas or opinions!

blimlim:
Thanks @jo-basevi! I think that sounds like a good plan. Happy to try and help rework some of the tests later!

Thanks for suggesting extending the test with extra inputs. I think that's a good way of going about it. I've added that in and added the config and resources.

Let me know if you have any suggested changes!

src/model_config_tests/models/accessesm1p5.py
```diff
@@ -74,14 +85,35 @@ def extract_checksums(
...
     ) -> dict[str, Any]:
         """Parse output file and create checksum using defined schema"""
-        if output_directory:
-            output_filename = output_directory / self.output_filename
-        else:
+
+        submodels = {
+            submodel["model"]: {"submodel_name": submodel["name"]}
+            for submodel in self.experiment.config["submodels"]
+        }
```
jo-basevi, line 92: As just the model types and names are used, could simplify this to a dictionary just mapping type to name, e.g. {"mom": "ocean", "um": "atmosphere"}. Though I think it is perfectly fine as-is so feel free to just resolve this comment

src/model_config_tests/models/um7.py
```diff
@@ -0,0 +1,42 @@
+"""Specific Mom5 postprocessing"""
+
+import re
+from collections import defaultdict
+from pathlib import Path
+
+FINAL_ABSOLUTE_NORM = "Final Absolute Norm"
+
+
+def um7_extract_norms(output_filename: Path, n_norms: int = 10) -> dict[str, list[any]]:
```
jo-basevi, line 10: Might be worth adding a global value for DEFAULT_N_NORMS = 10. Changing the default here is going to change the checksums for the amip configs - but that will be now caught with the expected checksums unit tests.

jo-basevi (approved):
Thanks @blimlim, changes look good to me. 

I wouldn't know on how many `Final Absolute Norms` to store, though storing more than one to avoid collisions seems a good idea. Ideally it would be great to have more descriptive checksums to give a starting point of what's changed, but having something to catch whether state has overall state changed is still good!

src/model_config_tests/models/accessesm1p5.py
blimlim: Good idea! I've added this in 402f2f8

src/model_config_tests/models/um7.py
blimlim: Added in 402f2f8!

blimlim:
Thanks @jo-basevi! I've just added in those changes. Are you happy to do a re-review?

src/model_config_tests/models/accessesm1p5.py
```diff
@@ -74,14 +85,33 @@ def extract_checksums(
...
+            output_filename = self.output_0 / submodels["um"] / "atm.fort6.pe0"
+            submodel_extract_checksums = um7_extract_norms
+
+        assert submodel_extract_checksums is not None, (
+            "Failed to find suitable submodel for checksum extraction."
+            "Required 'mom' or 'um' submodels not present in config.yaml "
+        )
+
+        if output_directory is not None:
+            output_filename = output_directory / output_filename
```
jo-basevi, line 110: Output directory is an option for when we want to check an output directory that isn't `output000`, e.g. the repro tests that runs the experiment twice sets `output_directory` to `/path/to/archive/output001` for the second model run. So we need the `output_filename` to be filename or a relative path? At the moment it is an absolute path in the output000 directory. Could also change the variable name for the absolute/full path to file from `output_filename`?
```suggestion
        if output_directory is not None:
            output_filepath = output_directory / output_filename
        else:
            output_filepath = self.output_0 / output_filename
```
Sorry I should've caught this on the first review!

blimlim: Thanks for catching this, and no need to apologise – I should have caught it while writing it!

I've added this in a72956c.

I've moved the section which sets the filename up into the initialisation. I noticed the reproducibility tests check that `self.output_file` exists before the checksums are extracted. In the earlier commits in this PR, self.output_file didn't necessarily match the file that the checksums were being extracted from, and so it was checking the wrong file. I've added some changes which should hopefully fix this!

Let me know if you have any suggestions!

src/model_config_tests/models/accessesm1p5.py
```diff
@@ -74,14 +94,27 @@ def extract_checksums(
...
+        if "mom" in self.submodels:
+            submodel_extract_checksums = mom5_extract_checksums
+        elif "um" in self.submodels:
+            # UM output is stored in submodel ouptut sub-directory
+            submodel_extract_checksums = um7_extract_norms
+
+        assert submodel_extract_checksums is not None, (
+            "Failed to find suitable submodel for checksum extraction."
+            "Required 'mom' or 'um' submodels not present in config.yaml "
+        )
```
jo-basevi, line 113: I'm not sure this assertion will be raised - as I think the code will exit in the `init` method if `self.output_file` is not defined. It might be worth adding this check, or similar, to the `init` method instead?

src/model_config_tests/models/accessesm1p6.py
```diff
@@ -1,11 +1,18 @@
...
         super().__init__(experiment)
 
-        self.output_filename = "access-esm1.6.out"
+        if "mom" in self.submodels:
+            self.output_filename = "access-esm1.6.out"
+        elif "um" in self.submodels:
+            # UM output is stored in submodel ouptut sub-directory
+            self.output_filename = Path(self.submodels["um"]) / "atm.fort6.pe0"
+
         self.output_file = self.output_0 / self.output_filename
```
jo-basevi, line 18: I think this is fine, but another option could be to have a `def set_output_files(self, model_stdout_file="access.out")` in `AccessEsm1p5`, that could be reused here with `self.set_output_files(model_stdout_file="access-esm1.6.out`)

jo-basevi (dismissed):
Thanks @blimlim, changes you added look good. Just a couple minor comments

src/model_config_tests/models/accessesm1p5.py
jo-basevi: Good catch! Yeah that check was a way to check if the model ended up running.. I'm currently looking into parsing out the exit code of jobs in a different issue to check if a payu exited with errors. But anyway still good to check that the file needed for checksums exists

src/model_config_tests/models/accessesm1p5.py
blimlim: Oh good point! Added in e6fa0e9

src/model_config_tests/models/accessesm1p6.py
blimlim: Good idea, I this is neater. I've added this in e6fa0e9

jo-basevi (approved):
Thanks @blimlim, all looks good to me

blimlim:
Thanks @jo-basevi for your help with this! Merging now

PR #147, opened 2025-03-28
Add new required input - a list of pytest markers for initial generation of checksums

### LINKED ISSUE #127: Cut down logic in generate-initial-checksums workflows
Each model repository has a generate checksum workflow (e.g. https://github.com/ACCESS-NRI/access-om2-configs/blob/main/.github/workflows/generate-initial-checksums.yml) which can be manually triggered to generate some initial checksums for a branch. Similarly to the CI workflows, most of the logic could possibly be moved to the reusable workflow `generate-checksums.yml` in this repository, so `generate-initial-checksums.yml` in the model configurations would just have one job to call the reusable workflow. This gives flexibility with updating the workflow, e.g. changing the `parse-ci-config` action for example, as not every model config repository would require updates.

### LINKED ISSUE #146: Hardcoded `checksum` marker no longer in use in `generate-initial-checksums`
See https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/14118889531/job/39556621043

## Background

We had hardcoded the `checksum` marker in `generate-initial-checksums` workflow - this isn't a good practice.
The markers changed in #137, which led to the workflows deselecting every test, and erroneously 'succeeding'.

CodeGat:
Closes #146
Closes #127

> [!IMPORTANT]
> This PR adds new required inputs to the `Generate Initial Checksums` workflow. Model Config Repos need to add this to the `generate-initial-checksums.yml` workflow in order for it to work. 

## Background

We hardcoded the `checksum` marker for Initial Checksum generation. When the marker changed, the workflow deselected all tests and succeeded vacuously.

Also, the caller workflows were quite large and did the same stuff, so we've moved all that logic inside model-config-tests

## The PR

* Add a required input to the generate initial checksums workflow - `test-markers`, rather than hardcoding `checksum`. Callers will have to update their workflows to use it. 
 * Move `config`, `log-inputs` jobs from caller to `model-config-tests
 * Updated `committed-checksum-tag` input to just the number rather than the whole tag, calling it `committed-checksum-tag-version` instead
 * Add a `committed-checksum-created-release` input to the caller so one can choose whether to create a release (in the case of `dev` branches.

### DIFF
--- a/.github/workflows/generate-checksums.yml
+++ b/.github/workflows/generate-checksums.yml
@@ -83,7 +83,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: |
-          echo '::notice::This deployment is using the following inputs: `config-branch-name`=`${{ inputs.config-branch-name }}`, `commit-checksums`=`${{ inputs.commit-checksums }}`, `committed-checksum-location`=`${{ inputs.committed-checksum-location }}`,`committed-checksum-tag-version`=`${{ needs.config.outputs.checksum-tag }}`, `test-markers`=`${{ inputs.test-markers }}`.'
+          echo '::notice::This deployment is using the following inputs: `config-branch-name`=`${{ inputs.config-branch-name }}`, `commit-checksums`=`${{ inputs.commit-checksums }}`, `committed-checksum-location`=`${{ inputs.committed-checksum-location }}`,`committed-checksum-tag-version`=`${{ needs.config.outputs.checksum-tag }}`, `committed-checksum-create-release`=`${{ inputs.committed-checksum-create-release }}`, `test-markers`=`${{ inputs.test-markers }}`.'
           echo '::notice::This deployment is using Payu Version ${{ needs.config.outputs.payu-version }} and Model Config Tests Version ${{ needs.config.outputs.model-config-tests-version }}'
 
   repro-ci:
--- a/.github/workflows/generate-checksums.yml
+++ b/.github/workflows/generate-checksums.yml
@@ -32,6 +32,10 @@ on:
         type: string
         required: true
         description: The payu module version used to create test virtual environment
+      test-markers:
+        type: string
+        required: true
+        description: A pytest-style string of pytest markers to run
     outputs:
       artifact-name:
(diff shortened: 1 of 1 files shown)


blimlim (approved):
Thanks @CodeGat. Looks good to me!

.github/workflows/generate-checksums.yml
```diff
@@ -43,15 +42,62 @@ on:
...
+        run: echo "checksum-tag=${{ inputs.config-branch-name }}-${{ inputs.committed-checksum-tag-version }}" >> $GITHUB_OUTPUT
+
+  log-inputs:
+    name: Log Inputs
+    needs:
+      - config
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo '::notice::This deployment is using the following inputs: `config-branch-name`=`${{ inputs.config-branch-name }}`, `commit-checksums`=`${{ inputs.commit-checksums }}`, `committed-checksum-location`=`${{ inputs.committed-checksum-location }}`,`committed-checksum-tag-version`=`${{ needs.config.outputs.checksum-tag }}`, `committed-checksum-create-release`=`${{ inputs.committed-checksum-create-release }}`, `test-markers`=`${{ inputs.test-markers }}`.'
```
jo-basevi, line 86: Should 
`committed-checksum-tag-version`=`${{ needs.config.outputs.checksum-tag }}`
be
`committed-checksum-tag-version`=`${{ inputs.committed-checksum-tag-version }}`?

CodeGat: True! Done in https://github.com/ACCESS-NRI/model-config-tests/pull/147/commits/f0c6065b0131fd6fa7d6b4a59a6b092031eb607e

CodeGat:
The test on the access-esm1.6-configs `dev-amip` branch was successful - see tag https://github.com/ACCESS-NRI/access-esm1.6-configs/releases/tag/dev-amip-0.1 and associated workflow run https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/14121994259. 

I'm going to merge this and ask for forgiveness later...

PR #149, opened 2025-04-02
Extra ACCESS-OM2 QA metadata tests

jo-basevi:
This PR adds QA tests for the metadata to check if model name is set as expected using the resolution of the target branch (https://github.com/ACCESS-NRI/access-om2-configs/issues/141), and that notes which have the conditions of use are set (https://github.com/ACCESS-NRI/access-om2-configs/issues/165).

Will leave this in draft until expected `model` names are confirmed.

### DIFF
--- a/src/model_config_tests/qa/test_access_om2_config.py
+++ b/src/model_config_tests/qa/test_access_om2_config.py
@@ -205,3 +205,14 @@ class TestAccessOM2:
             control_path=control_path,
             config=config,
         )
+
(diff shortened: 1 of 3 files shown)


jo-basevi:
Update the expected metadata notes based on review suggestions in https://github.com/ACCESS-NRI/access-om2-configs/pull/175

aidanheerdegen (approved):
LGTM

PR #162, opened 2025-06-04
accessom3: Change default run length to 6 hours

### LINKED ISSUE #161: Change default run length for ACCESS-OM3
We've encountered [an issue](https://github.com/ACCESS-NRI/access-om3-configs/pull/556#issuecomment-2939907442) where the default model test run length (3 hours) is not compatible with the timestep of some of the ACCESS-OM3 configurations. The team discussed on Zulip and decided that a 6 hour run length should be a good compromise between maximising divisors and minimising run length.

I'll follow up with a PR.

dougiesquire:
This PR changes the run length for ACCESS-OM3 repro tests to 6 hours

Closes #161

### DIFF
--- a/src/model_config_tests/models/accessom3.py
+++ b/src/model_config_tests/models/accessom3.py
@@ -8,11 +8,11 @@ import f90nml
 from netCDF4 import Dataset
 from payu.models.cesm_cmeps import Runconfig
 
-from model_config_tests.models.model import (
-    DEFAULT_RUNTIME_SECONDS,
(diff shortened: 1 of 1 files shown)


anton-seaice (approved):
Its not generally recommended to use code from the item under test within the test, but as it's just constants in this case, Ill let you off.

dougiesquire:
> Its not generally recommended to use code from the item under test within the test, but as it's just constants in this case, Ill let you off.

If you have another approach that will work here, let me know

PR #164, opened 2025-06-26
Use days for ESM runtime

### LINKED ISSUE #163: ESM1.6: allow for runtime given in days
With the latest [updates to payu](https://github.com/payu-org/payu/pull/585), the reproducibility tests are failing for ESM1.6. See [here for an example](https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/15889572690/job/44809354721).
```
RuntimeError: invalid runtime specified in config.yaml
```

I think the problem is that the ESM1.5/1.6 tests only support setting a runtime in units `years`, `months`, `seconds`:
https://github.com/ACCESS-NRI/model-config-tests/blob/2d000c586ef2340ffdb62f21b31527539147aeed/src/model_config_tests/models/accessesm1p5.py#L64-L70
and the standard 1 day test run is hence specified by setting the `seconds` option to `DAY_IN_SECONDS`.

However, the updated ESM1.6 payu driver ignores the `seconds` option in the `config.yaml`: 
https://github.com/payu-org/payu/blob/c2b975e961ed1748092ec2f0936e6933b44f7c58/payu/models/access_esm1p6.py#L190-L199

This results in a runtime of 0.

We could either allow the tests to set the runtime using units of days by adding a `days` argument to `set_model_runtime` https://github.com/ACCESS-NRI/model-config-tests/blob/2d000c586ef2340ffdb62f21b31527539147aeed/src/model_config_tests/models/accessesm1p5.py#L52-L54
Or we could modify the `payu` driver, allowing it to use runtimes given in units of `seconds`.

@jo-basevi and @anton-seaice, would you have a preference for either of the two options?

blimlim:
Closes #163. Sets the ESM1.5/1.6 runtime using smallest unit "days". This is for compatibility with the payu driver, which only uses the `years`, `months` and `days` options to set the runtime.

### DIFF
--- a/src/model_config_tests/models/accessesm1p5.py
+++ b/src/model_config_tests/models/accessesm1p5.py
@@ -61,12 +61,15 @@ class AccessEsm1p5(Model):
             seconds % DAY_IN_SECONDS == 0
         ), "Only days are supported in payu UM driver"
 
+        # ESM1.5/1.6 requires runtime in years, months, days
+        days = seconds / DAY_IN_SECONDS
+
         # Set runtime in config.yaml
         runtime_config = {
             "years": years,
             "months": months,
-            "days": 0,
-            "seconds": seconds,
+            "days": days,
+            "seconds": 0,
         }
         if "calendar" in doc:
             doc["calendar"]["runtime"] = runtime_config
--- a/tests/config_tests/test_test_bit_reproducibility.py
+++ b/tests/config_tests/test_test_bit_reproducibility.py
@@ -24,6 +24,7 @@ from model_config_tests.models.accessesm1p5 import (
 )
 from model_config_tests.models.accessom2 import DEFAULT_RUNTIME_SECONDS as OM2_RUNTIME
 from model_config_tests.models.accessom3 import DEFAULT_RUNTIME_SECONDS as OM3_RUNTIME
+from model_config_tests.util import DAY_IN_SECONDS
 
 
 def exp_test_helper_factory(*args, **kwargs):
@@ -457,8 +458,8 @@ def check_runtime(control_path, model_name):
         assert test_config["calendar"]["runtime"] == {
             "years": 0,
             "months": 0,
-            "days": 0,
-            "seconds": ESM_RUNTIME,
+            "days": ESM_RUNTIME / DAY_IN_SECONDS,
+            "seconds": 0,
         }
     elif model_name == "access-om2":
         with open(control_path / "accessom2.nml") as f:


src/model_config_tests/models/accessesm1p5.py
```diff
@@ -61,12 +61,15 @@ def set_model_runtime(
             seconds % DAY_IN_SECONDS == 0
         ), "Only days are supported in payu UM driver"
 
+        # ESM1.5/1.6 requires runtime in years, months, days
+        days = seconds / DAY_IN_SECONDS
```
blimlim, line 65: I've opted to keep the default runtime as `DEFAULT_RUNTIME_SECONDS`, converting it into days just before the config is written. This is to keep compatibility with other parts of the code which use `model.default_runtime_seconds` e.g. [here](https://github.com/ACCESS-NRI/model-config-tests/blob/2d000c586ef2340ffdb62f21b31527539147aeed/src/model_config_tests/config_tests/test_bit_reproducibility.py#L201)

anton-seaice:
I did a `pip install` from this repo, and ran `model-config-tests -m "repro and (not slow)" `

It failed with 

```
Experiment name is configured in config.yaml:  access-esm1.6-configs-exp_default_runtime
------------------------------------------------------------------------------------------------ Captured stderr setup ------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/603/as2285/.local/bin/payu", line 8, in <module>
    sys.exit(parse())
             ^^^^^^^
  File "/home/603/as2285/.local/lib/python3.12/site-packages/payu/cli.py", line 49, in parse
    run_cmd(**args)
  File "/home/603/as2285/.local/lib/python3.12/site-packages/payu/subcommands/setup_cmd.py", line 25, in runcmd
    expt = Experiment(lab, reproduce=reproduce, force=force,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/603/as2285/.local/lib/python3.12/site-packages/payu/experiment.py", line 96, in __init__
    self.init_models()
  File "/home/603/as2285/.local/lib/python3.12/site-packages/payu/experiment.py", line 165, in init_models
    ModelType = model_index[m_config['model']]
                ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
KeyError: 'access-esm1.6'
```

Which i think means the payu version is old. I think we need to update the dependencies in the `pyproject.toml` of this repo ?

If i manually do `pip install git+https://github.com/payu-org/payu` then the tests passes. 

Anyway - I guess i should make an new issue for this, as it's not this change that broke the dependency?

blimlim:
> Which i think means the payu version is old. I think we need to update the dependencies in the pyproject.toml of this repo ?
Good point, we might need to do a new payu release for it to work, as I think the ESM1.6 driver is only available in the dev version currently.

jo-basevi:
> Which i think means the payu version is old. I think we need to update the dependencies in the pyproject.toml of this repo ?

The CI builds the python virtual environments using the payu modules as a base for the dependencies (similar to the manual setup in the [readme](https://github.com/ACCESS-NRI/model-config-tests/?tab=readme-ov-file#how-to-run-pytests-manually-on-nci). So if `payu/dev` module would be how the tests are getting the latest payu version. 

To install directly from `model-config-tests` repo, then @blimlim is right that we will need to release a new payu version and that deploy a new version to PyPI.

src/model_config_tests/models/accessesm1p5.py
jo-basevi: Seconds is used to configure all the model run durations for all the repro tests currently (including the 1 and 2 day runs for restart repro and determinism). So I think adding the conversion here is the simplest.

src/model_config_tests/models/accessesm1p5.py
```diff
@@ -61,12 +61,15 @@ def set_model_runtime(
...
+
         # Set runtime in config.yaml
         runtime_config = {
             "years": years,
             "months": months,
-            "days": 0,
-            "seconds": seconds,
+            "days": days,
+            "seconds": 0,
```
jo-basevi, line 72: Could remove `seconds` fully from the config here, though having it set to 0 is also explicit.

PR #167, opened 2025-07-08
Add pytest error result to test-repro workflow

### LINKED ISSUE #166: Tests That Error Out Are Considered Succcessful Repro Checks in Comment
See https://github.com/ACCESS-NRI/access-esm1.6-configs/pull/154#issuecomment-3047304936, in which it states that it is a successful repro check, even though in the test check runs there was test errors https://github.com/ACCESS-NRI/access-esm1.6-configs/runs/45526335140. 

We need to check if any tests errored out, and have a suitable error message, rather than a success.

CodeGat:
Closes #166

## Background

Tests erroring out is not captured by the `test-repro` workflow. This means that if there are errors in the tests (not failures) they are treated as a not-failures in the comment, and therefore a success. 

More detail in the linked issue. 

This PR adds an `error` state to the `outputs.result` of `test=repro`, and comments that reflect this. 

## The PR

- **test-repro: Add new result output state `error` for when tests error out, rather than treating it as `pass`**
- **Update comments to reference a potential error state**

### DIFF
--- a/.github/workflows/config-comment-test.yml
+++ b/.github/workflows/config-comment-test.yml
@@ -232,6 +232,8 @@ jobs:
     steps:
       - name: Comment result
         env:
+          POTENTIAL_ERROR: |-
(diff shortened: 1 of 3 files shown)


CodeGat:
Thanks Jo :)

PR #173, opened 2025-07-29
Generate repro test errors when model does not run successfully

### LINKED ISSUE #170: Raise error in repro tests if model did not run
Currently, if a model failures to run, it is treated as a pytest failure rather than an error - though it would be good to change it to error so it's more clear whether a failure is a reproducibility error or payu is raising errors.

jo-basevi:
This PR:
   - Adds more stdout/error logs to exception messages so they will be displayed as part of annotations on the PR workflow logs (to make it easier to find out what went wrong) 
   - Moves checking whether the experiments ran without errors as part of setup in a fixture rather than within the test. This means any exceptions will be reported as a test setup error rather than a test failure. This will hopefully make it more clear when test failures are repro failures rather than the model did not run.

Closes #170

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,8 @@ tag_prefix = "v"
 parentdir_prefix = "model_config_tests-"
 
 [tool.coverage.run]
+patch = ["subprocess"]
 omit = [
-    "src/model_config_tests/_version.py"
+    "*/model_config_tests/_version.py",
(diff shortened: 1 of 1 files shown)


jo-basevi:
Rebased changes onto `main` and resolved conflicts.

CodeGat (approved):
We ran through this on a call, looks good!

Issue #175, opened 2025-07-30
Cleaning up stored repro test outputs


jo-basevi:
Test output from the reproducibility testing are stored on `/scratch` on Gadi, which is now taking up a bit of space.

Currently the files are stored under:
`${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.config-ref }}` where `inputs.config-ref` is:
- a branch name for scheduled tests and CI tests on PRs from `dev-*` -> `release-*` config branch
- a commit for manually triggered repro tests from a feature branch -> `dev-*` config branch (trigged by `!test repro`).

A useful thing about separate tests output by commit means that it allows comparing checksums between two intermediate commits in a PR. 

To routinely tidy up the test output, solutions could be:
- Remove model test output if there are no errors. Could check for the existence of the checksum file and if there's any reported errors in non historical checksum tests. Also checking for test errors rather than failures would require parsing the test report at some point so this clean up could be done later or last thing in the workflow.
- Store test output in `.../<branch or pr number>/<commit>` (note using `...` for `${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}`). Then once PR is closed remove `.../<branch or pr number>`.

jo-basevi:
Model test output is now removed if there are no errors as part of #176 

We could still remove outputs from PRs once a PR is closed.

PR #176, opened 2025-07-30
Clean up repro CI test outputs

jo-basevi:
This PR:
 - Adds a new input to `test-repro.yml` - `config-hash`. This is so all experiment directories will be set as
 `${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.config-ref }}/${{ inputs.config-hash }}` where `config-ref` is the branch or tag name and `config-hash` is the SHA commit hash. This is to make it easier to group commits to a particular branch in a PR (when eventually we set up automation to delete them when PR is closed) and also keep checksums on intermediate commits
 - After the test results are checked for errors (depending on https://github.com/ACCESS-NRI/model-config-tests/pull/173- this should indicate if the model did not run correctly),  remove any directories apart from `checksum`.
 
Contributes to #175 - this PR currently does not remove anything when config PRs are closed.

TODO:
- Testing?
- Double check the checksums file exists in artefact before deleting. 
- Could there be a repo setting to prevent deletion of outputs?
- Make the `rm -rf` slightly safer?

### DIFF
--- a/.github/workflows/test-repro.yml
+++ b/.github/workflows/test-repro.yml
@@ -2,10 +2,14 @@ name: Repro Checks
 on:
   workflow_call:
     inputs:
+      config-hash:
+        type: string
+        required: false
+        description: A SHA commit of the associated config branch to use for the reproducibility run
       config-ref:
         type: string
         required: true
-        description: A commit or tag on an associated config branch to use for the reproducibility run
+        description: A branch or tag on an associated config branch to use for the reproducibility run
       compared-config-ref:
         type: string
         required: false
@@ -100,6 +104,13 @@ jobs:
             ${{ secrets.SSH_HOST_DATA }}
           private-key: ${{ secrets.SSH_KEY }}
 
+      - name: Set experiment location
+        run: |
+          # Check to add commit hash to the experiment location if provided
+          if [[ -n "${{ inputs.config-hash }}" ]]; then
+            echo "EXPERIMENT_LOCATION=${{ env.EXPERIMENT_LOCATION }}/${{ inputs.config-hash }}" >> $GITHUB_ENV
+          fi
+
       - name: Run configuration
         id: run
         env:
@@ -120,7 +131,11 @@ jobs:
 
           # Setup a base experiment
           git clone ${{ github.event.repository.clone_url }} "${{ env.BASE_EXPERIMENT_LOCATION }}"
-          git -C "${{ env.BASE_EXPERIMENT_LOCATION }}" checkout ${{ inputs.config-ref }}
+          if [ -n "${{ inputs.config-hash }}" ]; then
+            git -C "${{ env.BASE_EXPERIMENT_LOCATION }}" checkout ${{ inputs.config-hash }}
+          else
+            git -C "${{ env.BASE_EXPERIMENT_LOCATION }}" checkout ${{ inputs.config-ref }}
+          fi
           cd "${{ env.BASE_EXPERIMENT_LOCATION }}"
 
           # Setup a compared checksum (if it exists)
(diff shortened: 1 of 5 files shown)


.github/workflows/test-repro.yml
```diff
@@ -67,7 +71,7 @@ jobs:
       artifact-url: ${{ steps.upload.outputs.artifact-url }}
       experiment-location: ${{ steps.run.outputs.experiment-location }}
     env:
-      EXPERIMENT_LOCATION: ${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.config-ref }}
+      EXPERIMENT_LOCATION: ${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.config-ref }}/${{ inputs.config-hash }}
```
CodeGat, line 74: One thing I'm wondering, perhaps for the future cleanup PR, is how would you determine what to delete with a `on.pull_request.types.closed` event? Would you also need to namespace the `EXPERIMENT_LOCATION` with a pull request number, so you can do something like `rm -rf '${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ github.event.pull_request.number }}/*`?

I suppose that would work for the pull request case, but we also run `test-repo.yml` with a `on.schedule` trigger, so that might not work, either...

jo-basevi: That's a good point.. For cleanup I was thinking of removing the source branch directory:
` rm -rf '${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ github.head_ref}}/*`
I thought there can't be two PRs open using the same source branch, and it shouldn't be an issue with scheduled checks as they use `release-` branches or tag and there wouldn't be PRs from a `release` branch.

*However* there are scheduled checks on some `dev-*` branches so that would cause conflicts.. So grouping by PR name would be the way to safely avoid conflicts with scheduled tests. So would need to pass in the pull request number or a namespace as an input to `test-repro.yml`? Should the naming also be changed for scheduled and initial checksums tests to make it more clear what was generated by what? E.g.
- PRs: `${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/pr6-dev-branch`
- Scheduled: `${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/schedule-dev-branch`
- Initial checksums: `${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/initial-dev-branch`

Or have `.../pr-<No>/<branch>/<commit>`, `.../schedule/<branch>/<commit>`, and `.../initial/<branch>/<commit>` to group by directories?

I'm not sure on the best way to clean up scheduled checks with branches, other than hope that there aren't many model failures so only checksums are being kept.

jo-basevi: I've added another input for `run_context` which would be `pr<number>` for pull request workflows, `schedule` for scheduled tests and `generate-checksums` for initial checksum workflows. Then used `run_context` to group branches/commits directories together in [cfa50f9](https://github.com/ACCESS-NRI/model-config-tests/pull/176/commits/cfa50f9817c52e5a1aaa4d40b391616faede67a3). It is maybe not the cleanest solution but should stop any conflicts between scheduled and PR test output..

.github/workflows/test-repro.yml
```diff
@@ -35,6 +39,10 @@ on:
         description: |
           Newline-separated paths for inclusion in the release artifact.
           Note that all paths given have 'env.EXPERIMENT_LOCATION/' prepended.
+      run_context:
```
Qian-HuiChen: I might be dummy but why do we use `run-context` in the previous few files but now we use `run_context` in `test-repro.yml`? How do they find each other?

jo-basevi: Added in 0a3d14f2a2e3c345939b21146e145aee9b1685a6

.github/workflows/test-repro.yml
```diff
@@ -71,7 +79,7 @@ jobs:
       artifact-url: ${{ steps.upload.outputs.artifact-url }}
       experiment-location: ${{ steps.run.outputs.experiment-location }}
     env:
-      EXPERIMENT_LOCATION: ${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.config-ref }}
+      EXPERIMENT_LOCATION: ${{ vars.EXPERIMENTS_LOCATION }}/${{ github.event.repository.name }}/${{ inputs.run-context }}/${{ inputs.config-ref }}/${{ inputs.config-hash }}
```
jo-basevi: Both the name of this path and the artefact later (`name=${{ github.event.repository.name }}-${{ inputs.run-context }}-${{ inputs.config-ref }}-${{ inputs.config-hash }}`) is quite long. Is this OK or should I shorten the config-hash?

CodeGat (approved):

PR #186, opened 2025-09-17
Fix parsing job ID in payu run submission stdout

### LINKED ISSUE #185: `ExpTestHelper` incorrectly parsing job ID
The `parse_run_id` method used by `ExpTestHelper` assumes that the run ID is the first line of the stdout from payu run. This no longer appears to be the case, e.g. see [this failed test](https://github.com/ACCESS-NRI/access-om3-configs/pull/758#issuecomment-3290760711) which, when run locally, gave:

```
Run Job ID: laboratory path:  /scratch/tm70/ds0092/tmp/test-model-repro/lab
-----Waiting for experiment exp_default_runtime to complete-----
Waiting for run job to finish. Job ID: laboratory path:  /scratch/tm70/ds0092/tmp/test-model-repro/lab
Error in experiment exp_default_runtime: Job ID laboratory path:  /scratch/tm70/ds0092/tmp/test-model-repro/lab not found in qstat output
```

I haven't looked closely at what caused the change in behaviour, but [this Payu PR](https://github.com/payu-org/payu/pull/558) seems probable. 

It looks like the run id is now the last line of the stdout (e.g. done [here](https://github.com/ACCESS-NRI/payu/blob/683afb9e3cf983a276d71afc0524acebf36986d5/payu/cli.py#L175-L178) in Payu).

@jo-basevi I'm happy to submit a PR if you agree with the above?

jo-basevi:
Payu in versions <= `1.1.7` prints the job ID as the first line of the payu run submission standard output, while in `payu/dev` it may be the last line. This PR re-uses a function that parses IDs from PBS stdout logs to parse only one ID from the printed output. 

Tested historical repro tests with an OM2 config with virtual environments built from the `payu/dev` module and the released `payu/1.1.7` module. Still running a full suite of repro tests an OM3 configuration for a sanity check 

Closes #185

### DIFF
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -345,10 +345,15 @@ def setup_exp(
 
 
 def parse_run_id(stdout: str) -> str:
-    """Parses the run ID from the subprocess stdout that submits payu run.
-    The run ID is the first line of the stdout."""
-    run_id = stdout.splitlines()[0]
-    return run_id
+    """Parses the Gadi PBS run ID from the subprocess stdout that submits payu
+    run"""
+    ids = parse_gadi_pbs_ids(stdout)
+    if len(ids) != 1:
+        raise RuntimeError(
+            "Expected 1 job ID in payu run submission, "
+            f"but found {len(ids)}. IDs: {ids}"
+        )
+    return ids[0]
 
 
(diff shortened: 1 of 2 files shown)


dougiesquire (approved):
Thanks @jo-basevi - looks great!

PR #190, opened 2025-11-28
Add repro payu setup check - this should pick up changes to manifests

jo-basevi:
This PR adds a "repro_payu_setup" test marker, which selects a test that runs "payu setup" with  "--reproduce" flag - This is equivalent to setting the following in `config.yaml`:
```yaml
manifests:
  reproduce:
    inputs: true
    exe: true
    restart: true
```
So setup will raise an error if any full hashes in the generated manifests do not match.

Tested this with running `model-config-tests -m repro_payu_setup` on a configuration with unchanged manifests, and with a changed restart manifest (removed values from a binhash and md5), the error is:

```
E           Failed: Failed to run payu setup with --reproduce. Error: Loading access-esm1p6/2025.11.002
E             Loading requirement: cice5/access-esm1.6-2025.11.001-6umbsy5 mom5/2025.05.000-tzlv4hw um7/2025.11.000-nwpmrxg
E           Run cannot reproduce: manifest manifests/restart.yaml is not correct
E
E           Full output: laboratory path:  /scratch/tm70/jb4202/tmp/test-model-repro/lab
E           binary path:  /scratch/tm70/jb4202/tmp/test-model-repro/lab/bin
E           input path:  /scratch/tm70/jb4202/tmp/test-model-repro/lab/input
E           work path:  /scratch/tm70/jb4202/tmp/test-model-repro/lab/work
E           archive path:  /scratch/tm70/jb4202/tmp/test-model-repro/lab/archive
E           Experiment name is configured in config.yaml:  dev-preindustrial+concentrations-repro_payu_setup
E           payu: Found modules in /opt/Modules/v4.3.0
E           Loading input manifest: manifests/input.yaml
E           Loading restart manifest: manifests/restart.yaml
E           Loading exe manifest: manifests/exe.yaml
E           Setting up atmosphere
E           Setting up ocean
E           Setting up ice
E           Setting up coupler
E           Checking exe, input and restart manifests
E           Manifest path: stored hash != calculated hash
E             work/atmosphere/restart_dump.astart: md5: f65a5a631c516907330e7f2d4479d5e != f65a5a631c516907330e7f2d4479d5e2
```

An issue is that Payu raises an error when restart manifests are empty, e.g. ACCESS-OM2: 
```
 Failed: Failed to run payu setup with --reproduce. Error: Restart manifest must exist and be populated if reproduce is configured to True
```
I wonder if payu could not error here and only error if there was no pre-existing manifest and a new restart manifest was generated? 

Related to #9

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Running all tests in the pytest suite on a configuration will likely fail as the
 - `repro_determinism_restart`: Determinism test that confirms repeated experiments with two consecutive runs give the same result.
 - `repro_restart`: Restart reproducibility test that confirms two short consecutive model runs give the same result as a longer single model run.
 - `repro_payu_setup`: Test payu setup reproducibility; fail if MD5 of any file in manifest is changed.
-- `manifests_unchanged`: Uses `git diff` to check for any unauthorized changes in manifest.
+- `manifests_unchanged`: Uses `git diff` to check manifests are up-to-date. If only fast hashes (e.g. `binhash`) are different, the manifests are reproducible, but `payu setup` may take longer to run as `md5` hashes need to be recalculated.  This test is not intended for tagged configurations.
 - `manifests`: A shortcut to run both `manifests_unchanged` and `repro_payu_setup`.
 - `slow`: Tests that are slow to run
 - `dev_config`: General configuration QA tests.
--- a/README.md
+++ b/README.md
@@ -90,9 +90,14 @@ Running all tests in the pytest suite on a configuration will likely fail as the
 - `repro_determinism`: Determinism test that confirms repeated model runs give the same result.
 - `repro_determinism_restart`: Determinism test that confirms repeated experiments with two consecutive runs give the same result.
 - `repro_restart`: Restart reproducibility test that confirms two short consecutive model runs give the same result as a longer single model run.
+- `repro_payu_setup`: Test payu setup reproducibility; fail if MD5 of any file in manifest is changed.
+- `manifests_unchanged`: Uses `git diff` to check for any unauthorized changes in manifest.
+- `manifests`: A shortcut to run both `manifests_unchanged` and `repro_payu_setup`.
 - `slow`: Tests that are slow to run
 - `dev_config`: General configuration QA tests.
 - `config`: Configuration QA tests for released branches. This includes the `dev_config` tests.
+
+
 
 There are also model-specific markers for configuration QA tests, e.g., `access_om2`, `access_esm1p5`, `access_om3` and `access_esm1p6`. For a list of all available markers,
 run:
--- a/src/model_config_tests/config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/config_tests/test_bit_reproducibility.py
@@ -333,7 +333,7 @@ class TestBitReproducibility:
 @pytest.mark.repro
 @pytest.mark.manifests
 @pytest.mark.repro_payu_setup
-def test_payu_setup_repro_flag(control_path, output_path):
+def test_repro_payu_setup(control_path, output_path):
     """
     Test payu setup with `--repro` flag which errors if md5 of any files in payu manifests are changed.
     """
--- a/src/model_config_tests/config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/config_tests/test_bit_reproducibility.py
@@ -9,7 +9,7 @@ from typing import Optional
 
 import pytest
 
-from model_config_tests.exp_test_helper import Experiments, ExpTestHelper
+from model_config_tests.exp_test_helper import Experiments, ExpTestHelper, setup_exp
 from model_config_tests.util import DAY_IN_SECONDS, HOUR_IN_SECONDS
 
 # Names of shared experiments
@@ -328,3 +328,15 @@ class TestBitReproducibility:
         )
 
         assert produced == expected
+
+
+@pytest.mark.repro_payu_setup
+def test_repro_payu_setup(control_path, output_path):
+    """
+    Test payu setup with --reproduce which errors if payu manifests full hashes are changed
+    """
+    experiment = setup_exp(control_path, output_path, exp_name="repro_payu_setup")
+    try:
+        experiment.setup_reproduce()
+    except Exception as error:
+        pytest.fail(f"{error}")
--- a/src/model_config_tests/config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/config_tests/test_bit_reproducibility.py
@@ -330,13 +330,30 @@ class TestBitReproducibility:
         assert produced == expected
 
 
+@pytest.mark.repro
+@pytest.mark.manifests
 @pytest.mark.repro_payu_setup
-def test_repro_payu_setup(control_path, output_path):
+def test_payu_setup_repro_flag(control_path, output_path):
     """
-    Test payu setup with `git diff` which errors if any files in payu manifests are changed.
+    Test payu setup with `--repro` flag which errors if md5 of any files in payu manifests are changed.
     """
     experiment = setup_exp(control_path, output_path, exp_name="repro_payu_setup")
     try:
         experiment.setup_reproduce()
+    except Exception as error:
+        pytest.fail(f"{error}")
+
+
+@pytest.mark.manifests
+@pytest.mark.manifests_unchanged
+def test_manifests_unchanged(control_path, output_path):
+    """
+    Test payu setup with `git diff` which errors if any files in payu manifests are changed.
+    """
+    experiment = setup_exp(
+        control_path, output_path, exp_name="setup_unchanged_manifests"
+    )
+    try:
+        experiment.setup_manifests_unchanged()
     except Exception as error:
         pytest.fail(f"{error}")
--- a/src/model_config_tests/config_tests/test_bit_reproducibility.py
+++ b/src/model_config_tests/config_tests/test_bit_reproducibility.py
@@ -333,7 +333,7 @@ class TestBitReproducibility:
 @pytest.mark.repro_payu_setup
 def test_repro_payu_setup(control_path, output_path):
     """
-    Test payu setup with --reproduce which errors if payu manifests full hashes are changed
+    Test payu setup with `git diff` which errors if any files in payu manifests are changed.
     """
     experiment = setup_exp(control_path, output_path, exp_name="repro_payu_setup")
     try:
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -82,6 +82,34 @@ class ExpTestHelper:
         """
         return self.model.output_exists()
 
+    def setup_reproduce(self):
+        """
+        Run payu setup command with --reproduce flag
+        """
+        owd = Path.cwd()
+        # Change to experiment directory and run.
+        os.chdir(self.control_path)
+
+        try:
+            setup_command = [
+                "payu",
+                "setup",
+                "--lab",
+                str(self.lab_path),
+                "--reproduce",
+            ]
+            print(f"Running payu setup command: {setup_command}")
+            result = sp.run(setup_command, capture_output=True, text=True)
+        finally:
+            # Change back to original working directory
+            os.chdir(owd)
+
+        if result.returncode != 0:
+            raise RuntimeError(
+                f"Failed to run payu setup with --reproduce. Error: {result.stderr}\n"
+                f"Full output: {result.stdout}"
+            )
+
     def setup_for_test_run(self):
         """
         Various config.yaml settings need to be modified in order to run in the
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -110,9 +110,22 @@ class ExpTestHelper:
             os.chdir(owd)
 
         if result.stdout != "":
-            raise RuntimeError(
-                f"Manifests have been modified. The modified files include: {result.stdout}.\n"
-            )
+            # Collect and display the top 10 lines of the diff for each modified file
+            files = result.stdout.strip().split("\n")
+            error_message = ""
+            for file in files:
+                error_message += f"Modifications are detected in {file}:\n"
+                diff_details = sp.run(
+                    ["git", "diff", f"{file}"],
+                    capture_output=True,
+                    text=True,
+                )
+                diff_lines = diff_details.stdout.splitlines()
+                top_lines = "\n".join(diff_lines[2:12])
+                if len(diff_lines) > 12:
+                    top_lines += "\n... (truncated)"
+                error_message += f"Changes are: \n {top_lines}\n"
+            raise RuntimeError(f"{error_message}")
 
     def setup_for_test_run(self):
         """
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -84,6 +84,34 @@ class ExpTestHelper:
 
     def setup_reproduce(self):
         """
+        Run payu setup with `--repro` flag to check if md5 hashes have changed in the manifests.
+        """
+        owd = Path.cwd()
+        # Change to experiment directory and run.
+        os.chdir(self.control_path)
+
+        try:
+            setup_command = [
+                "payu",
+                "setup",
+                "--lab",
+                str(self.lab_path),
+                "--reproduce",
+            ]
+            print(f"Running payu setup command: {setup_command}")
+            result = sp.run(setup_command, capture_output=True, text=True)
+        finally:
+            # Change back to original working directory
+            os.chdir(owd)
+
+        if result.returncode != 0:
+            raise RuntimeError(
+                f"Failed to run payu setup with --reproduce. Error: {result.stderr}\n"
+                f"Full output: {result.stdout}"
+            )
+
+    def setup_manifests_unchanged(self):
+        """
         Run payu setup command and check if manifests files have been changed with `git diff`.
         """
         owd = Path.cwd()
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -110,10 +110,27 @@ class ExpTestHelper:
         if result.returncode != 0:
             raise RuntimeError(
                 "Failed to run payu setup"
-                + (" with --reproduce" if reproduce else "")
+                + (" with --reproduce.\n" if reproduce else ".\n")
                 + f"{'='*10}STDOUT{'='*10}\n {result.stdout}\n"
                 f"{'='*10}STDERR{'='*10}\n {result.stderr}\n"
             )
+
+    def run_git_diff(self, path, extra_args=None):
+        """
+        Run git diff command on the given path and return the output.
+        """
+        command = ["git", "-C", str(path), "diff"] + extra_args if extra_args else []
+
+        result = sp.run(command, capture_output=True, text=True)
+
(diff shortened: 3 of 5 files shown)


jo-basevi:
Thinking about this more, it could just check if any `manifests/input.yaml`, `manifests/restart.yaml`, `manifests/exe.yaml` had changed.. 

Originally I thought it is not a good idea as it'll pick up changes to `binhashes` which can change even if the file is exactly the same. But I think it'll still be useful to update the `binhashes` on the configurations as it can lead to longer `payu setup` time in future, as it needs to recalculate the `md5` hashes.

Qian-HuiChen:
Thanks @jo-basevi  for the previous work and suggestions!
I change the `setup_reproduce` to run `payu setup` (no --reproduce flag)  and then check `git diff`. If the output is not empty, then returns an error. 

In this case, empty manifests will not raise an issue. 

Also test `model-config-tests -m repro_payu_setup` on unchanged manifests --> pass.
I changed manifests on git and run `model-config-tests -m repro_payu_setup` again, an error is raised as expected:
```
Manifests have been modified. The modified files include: restart.yaml
```

I am not sure if this is what we want. Happy to chat and change it.

src/model_config_tests/exp_test_helper.py
```diff
@@ -82,6 +82,51 @@ def has_run(self):
...
+        """
+        owd = Path.cwd()
+        # Change to experiment directory and run.
+        os.chdir(self.control_path)
+
+        try:
+            setup_command = ["payu", "setup", "--lab", str(self.lab_path)]
+            print(f"Running payu setup command: {setup_command}")
+            setup_result = sp.run(setup_command, capture_output=True, text=True)
+            if setup_result.returncode != 0 or "error" in setup_result.stderr.lower():
```
jo-basevi: If there are exceptions or sys.exit(1) should result in non-zero code. If there is a zero code, and there's an "error" in setup_result.stderr.lower(), I wonder if it would've coming from a warning or similar but not an error that would prevent the payu run. Probably fine to leave this, but I'm not 100% sure we should fail the test for that case.

src/model_config_tests/exp_test_helper.py
```diff
@@ -82,6 +82,51 @@ def has_run(self):
...
+            error_message = ""
+            for file in files:
+                error_message += f"Modifications are detected in {file}:\n"
+                diff_details = sp.run(
+                    ["git", "diff", f"{file}"],
+                    capture_output=True,
+                    text=True,
+                )
+                diff_lines = diff_details.stdout.splitlines()
+                top_lines = "\n".join(diff_lines[2:12])
```
jo-basevi, line 166: Could get an index error if len(diff_lines) < 12 here

I am wondering whether it should just show the full git diff here? If it did show the full error, it should show the all the names of modified files at the top. 

```
Changes detected in manifest file(s):
 - manifests/input.yml

If md5 hashes have changed, this indicates file contents being different.
If binhashes/paths have changed but md5's are the same, this will mean the configuration can reproduce the manifests but `payu setup` will take longer to run as it needs to re-calculate all the md5 hashes. 

Full git diff:
...
```

Also if you prefer the truncated sections, feel free to keep it too! Could start with truncated diff then if someone asks for the full diff, then it'll be easy enough to add it later

src/model_config_tests/exp_test_helper.py
```diff
@@ -82,6 +82,51 @@ def has_run(self):
...
+            os.chdir(owd)
+
+        if result.stdout != "":
+            # Collect and display the top 10 lines of the diff for each modified file
+            files = result.stdout.strip().split("\n")
+            error_message = ""
+            for file in files:
+                error_message += f"Modifications are detected in {file}:\n"
+                diff_details = sp.run(
+                    ["git", "diff", f"{file}"],
```
jo-basevi: This git diff may run from a different directory, so would be safer to use `git -C <path/to/repo>` e.g. 
```
"git diff -C {self.control_path} {file}"
```

jo-basevi (commented):
Thanks for adding this change in!!

Thinking about this more, I am wondering whether this should be two tests:

`test_setup_reproduce_manifests`
- Uses `payu setup --repro` (will need to fix payu failing on empty manifests first...). This will pick up if md5 hashes have changed (so the files are different).

`test_setup_unchanged_manifests`
- Catches if paths or binhashes have changed using the git diff logic you have added. If the `md5`'s are the same, it will just need the manifests to be updated with `payu setup`. This wouldn't be as large of an issue as the configurations are still using the same files but the initial `payu setup` would take longer as it needs to recalculate all the md5 hashes. 

For scheduled tests for tags, only `test_setup_reproduce_manifests` should run, as if it fails, it will indicate something on NCI has changed. 
But for PRs or scheduled tests for branches, both manifest tests should run so the manifests are always kept up to date.

src/model_config_tests/exp_test_helper.py
Qian-HuiChen: Thanks for pointing this out! Good point. It is implemented in the latest commit.

src/model_config_tests/exp_test_helper.py
Qian-HuiChen: >Could get an index error if len(diff_lines) < 12 here

I asked myself the same question before. So I tested the `top_lines = "\n".join(diff_lines[2:12])` when diff_lines are less than 12 rows. It reads in all lines without raising error. 
I even took an extreme test `top_lines = "\n".join(diff_lines[2:300])`. It reads in all output of `git diff`, no add-in empty lines and still didn't raise error. IndexError should not be an issue in this case.

I do agree that it should list all file names at the top, and then details of each file underneath. This is implemented in the latest commit.

src/model_config_tests/exp_test_helper.py
Qian-HuiChen: Good point. I agree that we shouldn't let exception fail the test. In the latest commit, the code will 
- raise error when payu setup returns non-zero code,
- print a warning when `error` is in stderr and payu setup exits with zero.

Qian-HuiChen:
>For scheduled tests for tags, only test_setup_reproduce_manifests should run, as if it fails, it will indicate something on NCI has changed.
But for PRs or scheduled tests for branches, both manifest tests should run so the manifests are always kept up to date.

I agree that it is good to split `test_setup_reproduce_manifests` and `test_setup_unchanged_manifests` for different usages and triggered differently. I have opened a[ payu issue](https://github.com/payu-org/payu/issues/704) to solve the empty manifest handling issue first, and then come back to modify the model-config-tests.

src/model_config_tests/exp_test_helper.py
jo-basevi: Huh I did not know that Python list slicing did not raise IndexErrors! (e.g. `["a", "b"][2:12]` return []`)

src/model_config_tests/exp_test_helper.py
```diff
@@ -82,6 +82,86 @@ def has_run(self):
...
+            result = sp.run(setup_command, capture_output=True, text=True)
+        finally:
+            # Change back to original working directory
+            os.chdir(owd)
+
+        if result.returncode != 0:
+            raise RuntimeError(
+                f"Failed to run payu setup with --reproduce. Error: {result.stderr}\n"
+                f"Full output: {result.stdout}"
+            )
```
jo-basevi: Could move the setup code to one function that has an optional reproduce flag as both tests have similar setup?

```suggestion
    def setup(self, reproduce=False):
        """
        Run payu setup command. If reproduce is True, run with --repro flag to
        check if md5 hashes have changed in the manifests.
        """
        owd = Path.cwd()
        # Change to experiment directory
        os.chdir(self.control_path)
  
        try:
            setup_command = ["payu", "setup", "--lab", str(self.lab_path)]
            if reproduce:
                setup_command.append("--reproduce")
            print(f"Running payu setup command: {setup_command}")
            result = sp.run(setup_command, capture_output=True, text=True)
        finally:
            # Change back to original working directory
            os.chdir(owd)
  
        if result.returncode != 0:
            raise RuntimeError(
                f"Failed to run payu setup" + (f" with --reproduce" if reproduce else "")
                f"{'='*10}STDOUT{'='*10}\n{setup_result.stdout}\n"
                f"{'='*10}STDERR{'='*10}\n{setup_result.stderr}\n"
            )
```

tests/test_exp_test_helper.py
```diff
@@ -556,3 +556,74 @@ def test_experiments_check_experiment_error(tmp_path):
...
+    mock_run.return_value = mock_result
+
+    with pytest.raises(RuntimeError) as excinfo:
+        exp.setup_manifests_unchanged()
+
+        assert "Error during payu setup." in str(excinfo.value)
+        assert f"{'='*10}STDOUT{'='*10}\n {mock_result.stdout}\n" in str(excinfo.value)
+
+        # assert returning to the original work directory
+        assert Path.cwd() == exp.control_path
```
jo-basevi:
```suggestion
    # Store original current working directory
    owd = Path.cwd()
   
    with pytest.raises(RuntimeError) as excinfo:
        exp.setup_manifests_unchanged()

    assert "Error during payu setup." in str(excinfo.value)
    assert f"{'='*10}STDOUT{'='*10}\n {mock_result.stdout}\n" in str(excinfo.value)

    # assert returning to the original work directory
    assert Path.cwd() == owd
```

The `assert`s need to be moved outside of the `with pytest.raises(RuntimeError) as excinfo` as otherwise they are never reached if `exp.setup_manifests_unchanged()` raises an error.

jo-basevi: This applies to the other 2 tests too

tests/test_exp_test_helper.py
```diff
@@ -556,3 +556,74 @@ def test_experiments_check_experiment_error(tmp_path):
...
+
+
+@patch("subprocess.run")
+def test_setup_manifests_unchanged_show_changes(mock_run, exp):
+    """Test that when manifests are changed, the `git diff` results are printed to stdout"""
+    # Mock the `payu setup` succeed first
+    setup_success = MagicMock(returncode=0, stdout="Payu setup succeeded")
+
+    # Then mock the `git diff --name-only` to show which files are changed
+    git_diff_name_only = MagicMock(returncode=1, stdout="manifests/input.yaml")
```
jo-basevi: Should we error if returncode != 1?

tests/test_exp_test_helper.py
```diff
@@ -556,3 +556,74 @@ def test_experiments_check_experiment_error(tmp_path):
...
+
+    # Mock the `git diff` to show the detailed changes in the file
+    git_diff_output = (
+        "--- a/manifests/input.yaml\n+++ b/manifests/input.yaml\n+new line\n-old line"
+    )
+    git_diff_run = MagicMock(returncode=0, stdout=git_diff_output)
+
+    # Run these mocks in sequence
+    mock_run.side_effect = [setup_success, git_diff_name_only, git_diff_run]
```
jo-basevi, line 634: Nice!

jo-basevi (commented):
Thanks for adding these changes in!

jo-basevi:
It could be good add the new markers to the list of markers in the README.md:

 https://github.com/ACCESS-NRI/model-config-tests?tab=readme-ov-file#selecting-tests-using-markers

tests/test_exp_test_helper.py
Qian-HuiChen: Good point. I double check [git command](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-diffdrivertrustExitCode). `git diff` returns `0` no matter whether changes are detected.

It turns out that previous code [f277c4c](https://github.com/ACCESS-NRI/model-config-tests/pull/190/commits/f277c4c15301a2d0ca939e274b0549d0d6ec1a60) did not check returncode, but only checks if stdout is empty or not. Therefore, even though the mock says returncode is `1`, payu repro check didn't error.

I upgraded the `setup_manifests_unchanged` to check `git diff` returncode, error out if not zero, otherwise continue to check if stdout is empty.

tests/test_exp_test_helper.py
Qian-HuiChen: Thanks 😄

tests/test_exp_test_helper.py
Qian-HuiChen: Oh sorry I didn't realize this. I corrected this in the latest commit.

src/model_config_tests/exp_test_helper.py
Qian-HuiChen: This is a great idea to keep the code clean! I implemented this in the latest commit.

Qian-HuiChen:
Thanks @jo-basevi for reminding me of updating readme :)

src/model_config_tests/exp_test_helper.py
jo-basevi: Nice!

src/model_config_tests/exp_test_helper.py
```diff
@@ -82,6 +82,87 @@ def has_run(self):
...
+        self.setup(reproduce=True)
+
+    def setup_manifests_unchanged(self):
+        """
+        Run payu setup command and check if manifests files have been changed with `git diff`.
+        """
+        self.setup(reproduce=False)
+
+        result = sp.run(
+            ["git", "diff", "--name-only", str(self.control_path / "manifests/")],
```
jo-basevi: Probably best to use `-C <ctrl_path>` here as well so it is as-if was running the command from the control directory. 

Could save a base diff command to a variable to reuse?
`base_diff_cmd = ["git", "-C", str(self.control_path), "diff"]`

Then the subsequent commands could be:
`sp.run(base_diff_cmd + ["--name-only", "manifests/"])`
`sp.run(base_diff_cmd + [file])`

OR put the git diff into a function that runs the command and which will checks the return code for both git diff commands? It could live outside of the experiment test helper class too as it will only need the repository path, so doesn't need the entire `self` object? 

e.g.

```python
def run_git_diff(repo_path, extra_cmd_flags):
    git_diff_cmd = ["git", "-C", str(repo_path), "diff"] + extra_cmd_flags
    result = sp.run(
            git_diff_cmd,
            capture_output=True,
            text=True,
        )

      if result.returncode != 0:
          raise RuntimeError(
              f"Git command failed with exit code {result.returncode}.\n"
              # Display git_diff_cmd that was run for debugging
              f"Command: " + " ".join(git_diff_cmd) + "\n"
              f"{'='*10}STDOUT{'='*10}\n {result.stdout}\n"
              f"{'='*10}STDERR{'='*10}\n {result.stderr}\n"
          )
    return result.stdout


# Called with 
diff_output = run_git_diff(self.control_path, ["--name-only", "manifests/"])
#...
diff_details = run_git_diff(self.control_path, [file])
```

tests/test_exp_test_helper.py
```diff
@@ -556,3 +556,83 @@ def test_experiments_check_experiment_error(tmp_path):
...
+        exp.setup_reproduce()
+
+        assert (
+            f"Failed to run payu setup with --reproduce. Error: {mock_result.stderr}"
+            in str(excinfo.value)
+        )
+        assert f"Full output: {mock_result.stdout}" in str(excinfo.value)
+
+        # assert returning to the original work directory
+        assert Path.cwd() == exp.control_path
```
jo-basevi: Same issue here with the asserts never being reached

tests/test_exp_test_helper.py
```diff
@@ -556,3 +556,83 @@ def test_experiments_check_experiment_error(tmp_path):
...
+    setup_success = MagicMock(returncode=0, stdout="Payu setup succeeded")
+
+    top_lines = "+new line\n-old line"
+    diff_file = "manifests/input.yaml"
+    # Then mock the `git diff --name-only` to show which files are changed
+    git_diff_name_only = MagicMock(returncode=0, stdout=diff_file)
+
+    # Mock the `git diff` to show the detailed changes in the file
+    git_diff_run = MagicMock(
+        returncode=0, stdout=(f"--- a/{diff_file}\n+++ b/{diff_file}\n" + top_lines)
```
jo-basevi: stdout would be more similar to:
"""
diff --git a/manifests/input.yaml b/manifests/input.yaml
index abc123...zyx789 100644
--- a/manifests/input.yaml 
+++ b/manifests/input.yaml 
+new line
-old line
"""

As I think the top two lines get removed with `top_lines = "\n".join(diff_lines[2:12])`

src/model_config_tests/config_tests/test_bit_reproducibility.py
```diff
@@ -328,3 +328,32 @@ def test_repro_determinism_restart(
         )
 
         assert produced == expected
+
+
+@pytest.mark.repro
+@pytest.mark.manifests
+@pytest.mark.repro_payu_setup
+def test_payu_setup_repro_flag(control_path, output_path):
```
jo-basevi: Previously I had to make a change to the other repro tests, to have the test name match the specific marker, so might be good to rename this to `test_repro_payu_setup`

```suggestion
def test_repro_payu_setup(control_path, output_path):
```

README.md
```diff
@@ -90,10 +90,15 @@ Running all tests in the pytest suite on a configuration will likely fail as the
 - `repro_determinism`: Determinism test that confirms repeated model runs give the same result.
 - `repro_determinism_restart`: Determinism test that confirms repeated experiments with two consecutive runs give the same result.
 - `repro_restart`: Restart reproducibility test that confirms two short consecutive model runs give the same result as a longer single model run.
+- `repro_payu_setup`: Test payu setup reproducibility; fail if MD5 of any file in manifest is changed.
+- `manifests_unchanged`: Uses `git diff` to check for any unauthorized changes in manifest.
```
jo-basevi: It might be worth adding a description of what this test does and when to use it. Does the following explanation make sense?

```suggestion
- `manifests_unchanged`: Uses `git diff` to check manifests are up-to-date. If only fast hashes (e.g. `binhash`) are different, this means the manifests reproduce but `payu setup` can take longer to run as `md5` hashes must be recalculated.  This test is not intended to be run for tagged configurations. 
```

jo-basevi (commented):
Changes look good! Just a couple very minor comments

README.md
Qian-HuiChen: Thanks for the detailed descriptions. Sounds very clear.

src/model_config_tests/config_tests/test_bit_reproducibility.py
Qian-HuiChen: Make sense. I change the name in the latest commit.

tests/test_exp_test_helper.py
Qian-HuiChen: I see. I improved the mock git_diff_run in the latest commit.

tests/test_exp_test_helper.py
Qian-HuiChen: Thanks for spotting this! I wrote the real-functioning test in the latest commit.

src/model_config_tests/exp_test_helper.py
Qian-HuiChen: Sounds good! I make a new function `run_git_diff` in the latest commit.

jo-basevi (commented):
Thanks @Qian-HuiChen, changes look good to me! I've run a couple tests and it seems to run as expected

I'm not able to approve this PR as I opened the PR initially. As there doesn't seem to be a way to change authorship, it might make sense to close this PR and re-open a new PR, which I can then approve?

Qian-HuiChen (approved):
Since Jo cannot approve it on her side, I will approve myself haha

Qian-HuiChen:
Turns out my approval cannot allow merging because I pushed the last commit. I re-opened this PR #208.

PR #198, opened 2026-03-02
Adding PAYU_PBSNODES_CACHE variable into repro CI.

### LINKED ISSUE #196: Set PAYU_PBSNODES_CACHE in Repro CI tests
Currently payu caches PBS node info to `$HOME/.cache/pbs/`. To avoid using the service user's home directory in tests, we can set `PAYU_PBSNODES_CACHE` environment variable to a custom path. 

This could be set in the ssh session that runs the tests (which run payu):

https://github.com/ACCESS-NRI/model-config-tests/blob/31685c167c305b26642a22f00796dec7d8725551/.github/workflows/test-repro.yml#L107-L114

A central location could be `${{ vars.EXPERIMENTS_LOCATION }}/.cache/pbs/`? This is the top-level directory  which contains all the test files - so it'll be reused across all repro tests.

Qian-HuiChen:
Closes #196 .
Reference [PBS timeout issue](https://github.com/payu-org/payu/issues/688)


jo-basevi (approved):
LGTM!

PR #205, opened 2026-03-16
ACCESS-OM2 QA checks: Only expect ACCESS-OM2-BGC for old BGC configs

dougiesquire:
Sorry @jo-basevi, I rushed this the first time and missed the module and repo name checks:

The behaviour should be:
- if "bgc" is in the config name: use `ACCESS-OM2-BGC` (now-deprecated)
- if "wombat" is in the config name: use `ACCESS-OM2`


jo-basevi (approved):
Thanks - LGTM

PR #208, opened 2026-04-16
Add repro payu setup check

Qian-HuiChen:
This PR is migrated from #190 (opened by @jo-basevi and continued works by @Qian-HuiChen ). It is re-opened here so @jo-basevi  can approve the commits.


jo-basevi (approved):
Thanks! Approved :)

PR #212, opened 2026-04-22
Remove test configurations in `tests/resources`

### LINKED ISSUE #117: Remove test configurations in `tests/resources`?
As [suggested by @aidanheerdegen](https://github.com/ACCESS-NRI/model-config-tests/pull/110#discussion_r1970816071), we could remove the test configurations in `tests/resources` and instead clone a specific tag from the relevant config at test time. One disadvantage of doing this is that the tests then would require internet access.

Qian-HuiChen:
Remove all configurations stored in `tests/resources` which are used in unit testing.

Instead, on-demand cloning is called via `git clone` during unit test execution and cleaned up after the unit tests. A fixed commit hash is currently hard-coded in `tests/common.py` to ensure reproducibility and version control.

Closes #117.


tests/common.py
```diff
@@ -1,5 +1,115 @@
...
+    "om2-1deg": {
+        "repo_url": "https://github.com/ACCESS-NRI/access-om2-configs.git",
+        "branch": "release-1deg_jra55_ryf",
+        "file_list": [
+            "manifests/",
+            "config.yaml",
+            "accessom2.nml",
+            "metadata.yaml",
+            "LICENSE",
+        ],
```
jo-basevi: I think it'll be ok to clone all the files in the configurations to make it simpler. I think a subset was only to saved to the repository's test resources previously to avoid saving too many files.

jo-basevi (dismissed):
Thanks for implementing this @Qian-HuiChen. It is nice to see so many deleted files :) 

As the cloned configurations are used in several tests, an option could be to have a session-scoped fixture that creates a cache of the cloned configurations. Then a test-scoped fixture that copies the clone to a tmp directory, and then cleans up the directory after. This might run faster than cloning for every test.

tests/common.py
Qian-HuiChen: Sounds good! I make clone_config a fixture and it clone all files now. Just found that ESM1.5 has their checksum updated, so I updated the checksum and output as well.

tests/config_tests/qa/test_test_access_esm1p6_config.py
```diff
@@ -1,26 +1,28 @@
 # import shlex
+# import shutil
 # import subprocess
 
-# from tests.common import RESOURCES_DIR
-
 
 # TODO: Add back when we have a sufficient test config released
-def test_test_access_esm1p6_config_release_release_preindustrial():
+def test_test_access_esm1p6_config_release_release_preindustrial(
```
jo-basevi: Could uncomment this as there's probably a dev configuration that passes the access_esm1p6 qa checks now

tests/conftest.py
```diff
@@ -0,0 +1,41 @@
...
+        if config_name not in cache:
+            # Clone the remote repository into the cache directory
+            dest = cache_root / config_name
+            dest.mkdir(parents=True, exist_ok=True)
+            clone_config_repo(config_name, dest)
+            cache[config_name] = dest
+        return CLONE_CONFIGS[config_name]["branch"], cache[config_name]
+
+    return _get_config_cache
+
```
jo-basevi: I've noticed you have been removing tmp_path after each test (with `shutil.rmtree(tmp_path)`)

you could do something similar here with [`yield`](https://docs.pytest.org/en/stable/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization), e.g.

```python
yield _get_config_cache

# Runs at the end of session
shutil.rmtree(cache_root)
```

Maybe could use `yield` to remove tmp_path, if it exists, in the isolated_config fixture below which saves you having to add it for each test?

jo-basevi (dismissed):
The fixture implementation looks neat! The latest commit does speed up the tests :) On my local laptop, the tests in [9595035](https://github.com/ACCESS-NRI/model-config-tests/pull/212/commits/9595035cbac4d3b2b74ea59f4b854c64bf58a5cc) took about 90s to run and now in [7a19ac6](https://github.com/ACCESS-NRI/model-config-tests/pull/212/commits/7a19ac643b9a6ee2c1b208bcabb192ec4736b1a2), its about 45s 

Both my comments on code are minor and can be left, but it would good to reduce the size of the test esm1.6's `access.out` file if possible, as it's a large file to add to the repository.

tests/conftest.py
Qian-HuiChen: Sounds good. I implemented this in [936b19e](https://github.com/ACCESS-NRI/model-config-tests/pull/212/commits/936b19e0dd83179af306d872b2bc96c58f3e9ec5)

tests/config_tests/qa/test_test_access_esm1p6_config.py
Qian-HuiChen: I uncommented this and confirmed that the test still passed ([7afe6ef](https://github.com/ACCESS-NRI/model-config-tests/pull/212/commits/7afe6ef06692e05b682b01001745744ed2330011)).

Qian-HuiChen:
I tried to trim down `esm1.6/output000/access.out` in commit [7afe6ef](https://github.com/ACCESS-NRI/model-config-tests/pull/212/commits/7afe6ef06692e05b682b01001745744ed2330011).

jo-basevi (dismissed):
Thanks @Qian-HuiChen, looks good to me :)

jo-basevi (approved):

PR #214, opened 2026-05-04
Add sync:base_path check in QA test.

### LINKED ISSUE #213: Update QA tests to check `base_path` instead of `path`
Currently QA test check the `sync` is not enabled and `path` is not set ([ref codes](https://github.com/ACCESS-NRI/model-config-tests/blob/main/src/model_config_tests/config_tests/qa/test_config.py#L222-L232)).

Payu 1.3.0 updated `base_path` as a safer alternative of `path` to reduce the risk of overwriting outputs. To align with this update, QA checks should be updated to check on using `base_path`.

Ref [payu issue #724](https://github.com/payu-org/payu/issues/724).

Qian-HuiChen:
This PR updates the QA test to check `base_path` is not set in sync in config.yaml, instead of checking `path`.

Closes #213.

### DIFF
--- a/src/model_config_tests/config_tests/qa/test_config.py
+++ b/src/model_config_tests/config_tests/qa/test_config.py
@@ -233,11 +233,10 @@ class TestConfig:
             ), "Sync path to remote archive should not be set"
 
     def test_sync_path_is_none(self, config):
-        if "sync" in config:
-            if "path" in config["sync"]:
-                assert (
-                    config["sync"]["path"] is None
-                ), "Sync paths should be None if it is set"
+        if "sync" in config and "path" in config["sync"]:
+            assert (
+                config["sync"]["path"] is None
+            ), "Sync paths should be None if it is set"
 
     def test_experiment_name_is_not_defined(self, config):
         assert "experiment" not in config, (
--- a/src/model_config_tests/config_tests/qa/test_config.py
+++ b/src/model_config_tests/config_tests/qa/test_config.py
@@ -225,11 +225,19 @@ class TestConfig:
                 "enable"
             ], "Sync to remote archive should not be enabled"
 
-    def test_sync_path_is_not_set(self, config):
+    def test_sync_base_path_is_not_set(self, config):
         if "sync" in config:
             assert not (
-                "path" in config["sync"] and config["sync"]["path"] is not None
+                "base_path" in config["sync"]
+                and config["sync"]["base_path"] is not None
             ), "Sync path to remote archive should not be set"
+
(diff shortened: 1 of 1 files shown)


src/model_config_tests/config_tests/qa/test_config.py
```diff
@@ -225,12 +225,20 @@ def test_sync_is_not_enabled(self, config):
...
         if "sync" in config:
             assert not (
-                "path" in config["sync"] and config["sync"]["path"] is not None
+                "base_path" in config["sync"]
+                and config["sync"]["base_path"] is not None
             ), "Sync path to remote archive should not be set"
 
+    def test_sync_path_is_none(self, config):
+        if "sync" in config:
+            if "path" in config["sync"]:
```
atteggiani: This is equivalent to:
```suggestion
        if "sync" in config and "path" in config["sync"]:
```

atteggiani (commented):
This looks good @Qian-HuiChen 

Just a small simplification

src/model_config_tests/config_tests/qa/test_config.py
Qian-HuiChen: Thanks! Initially I was not sure if this will return a key error if "sync" is not in config. Your comments gave me the confidence to test it and turns out: it doesn't error out if sync is not a key. I implemented this in the latest commit.

atteggiani: Yes it doesn't. The conditions are executed in order of appearance from left to right and the `if` clause immediately exists if `false` gets hit.
So if the `"sync" in config` portion is false, the `"path" in config["sync"]` part will never be executed.

atteggiani (approved):

src/model_config_tests/config_tests/qa/test_config.py
```diff
@@ -225,12 +225,19 @@ def test_sync_is_not_enabled(self, config):
...
-                "path" in config["sync"] and config["sync"]["path"] is not None
+                "base_path" in config["sync"]
+                and config["sync"]["base_path"] is not None
             ), "Sync path to remote archive should not be set"
 
+    def test_sync_path_is_none(self, config):
+        if "sync" in config and "path" in config["sync"]:
+            assert (
+                config["sync"]["path"] is None
+            ), "Sync paths should be None if it is set"
```
aidanheerdegen, line 239: We now prefer to use `base_path`, so I think the test should not allow `config["sync"]["path"]` at all, because we do not want to suggest to users that this is the preferred option.

Sorry for commenting on a closed PR!

PR #219, opened 2026-05-19
CI: Update Actions To Use Node 24

CodeGat:
See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

## The PR

- **Update actions/checkout to v6**
- **Update actions/setup-python to v6**
- **Update EnricoMi/publish-unit-test-result-action to use non-composite action, updated to v2.23.0** - see https://github.com/EnricoMi/publish-unit-test-result-action/tree/v2#running-as-a-composite-action

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Setup Python 3.11
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: 3.11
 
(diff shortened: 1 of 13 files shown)


.github/workflows/CD.yml
jo-basevi, line 44:
```suggestion
      - uses: actions/download-artifact@v7
```

jo-basevi, line 29:
```suggestion
        uses: actions/upload-artifact@v7
```

Qian-HuiChen (approved):
Looks good to me. Double checked the action version hash as well.

jo-basevi (commented):
There's a few workflows with upload/download artifact at `v4`:

PR #224, opened 2026-06-02
For PR workflows, use `github.event.pull_request.head.sha` rather than `github.sha`

CodeGat:
## Background

See error https://github.com/ACCESS-NRI/access-esm1.6-configs/actions/runs/26791433362/job/78987492391#step:4:94

The PR workflow uses `github.sha` for `test-repro`s `inputs.config-hash`, but for pull request events, this resolves to the merge commit (a GitHub-specific creation), which isn't accessible to cloned repos. 

Rather than that, we should use `github.event.pull_request.head.sha`, which resolves to the `HEAD` of the PR branch. 

## The PR

* In the PR workflow, use `github.event.pull_request.head.sha`.


jo-basevi (approved):
LGTM

PR #225, opened 2026-06-02
Update README about payu minimum version.

### LINKED ISSUE #165: Some tests require not yet release version of payu
As [noted](https://github.com/ACCESS-NRI/model-config-tests/pull/164#issuecomment-3017300739) , there are tests in this repo which rely on the access-esm1.6 driver in payu, however this was added to payu after the latest release.

Therefore if a user does `pip install` from the model-config-tests repo, the access-esm1.6 test won't run unless the user manually updates the payu version from `master` branch in https://github.com/payu-org/payu

When run through the CI, `model-config-tests` use [installed payu on NCI](https://github.com/ACCESS-NRI/model-config-tests/pull/164#issuecomment-3017300739). So this is mostly an issue for user installs. 

We could
- do a payu release and update the dependency [version](https://github.com/ACCESS-NRI/model-config-tests/blob/2d000c586ef2340ffdb62f21b31527539147aeed/pyproject.toml#L24)
- change the dependency to pull from github ?
- note this in the docs?
- other?

Qian-HuiChen:
This PR updates the README by adding instructions about `payu_minimum_version` requirements.

Closes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,19 @@ Code from these pytests is adapted from COSIMAS's ACCESS-OM2's [bit reproducibil
 
 ### How to run pytests manually on NCI
 
-1. Load payu module - this provides the dependencies needed to run the model
+1. Load payu module - this provides the dependencies needed to run the model.
 
     ```sh
     module use /g/data/vk83/modules
     module load payu
+    ```
+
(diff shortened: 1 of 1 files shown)


jo-basevi (approved):
Thanks, LGTM

PR #232, opened 2026-06-29
Print payu run error message; Update resource for unit test

### LINKED ISSUE #229: Add stdout/stderr logs for failed payu run submission
As raised in https://github.com/ACCESS-NRI/access-om3-configs/pull/1432#issuecomment-4741169389, it's difficult to see what the issue when submitting a `payu run` job from the logs.

We have previously added the stdout/stderr from running `payu setup` here:

https://github.com/ACCESS-NRI/model-config-tests/blob/ee4cb743816648753c36e15e6a630bedc2859853/src/model_config_tests/exp_test_helper.py#L227-L239

This will just need something similar for `payu run`. As `sp.CalledProcessError` already adds stdout/stderr fields logs to the error objects, this should be straight-forward to add here:

https://github.com/ACCESS-NRI/model-config-tests/blob/ee4cb743816648753c36e15e6a630bedc2859853/src/model_config_tests/exp_test_helper.py#L258-L259

Qian-HuiChen:
This PR fixes two issues:
1. Print error message for `payu run` in subprocess. Closes #229.
2. Update the OM2 and OM3 resources for unit tests, since the previous branches are removed in the [OM2](https://github.com/ACCESS-NRI/access-om2-configs/issues/352) and [OM3](https://github.com/ACCESS-NRI/access-om3-configs/pull/1440) configs.

### DIFF
--- a/src/model_config_tests/exp_test_helper.py
+++ b/src/model_config_tests/exp_test_helper.py
@@ -223,6 +223,7 @@ but `payu setup` will take longer to run as it needs to re-calculate all the md5
             # Change to experiment directory and run.
             os.chdir(self.control_path)
 
+            # Run payu setup command
             print("Running payu setup")
             result = sp.run(
                 ["payu", "setup", "--lab", str(self.lab_path)],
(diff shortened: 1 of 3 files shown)


jo-basevi (approved):
Thanks for adding this in! Looks good to me

PR #233, opened 2026-07-14
Separate test_test_repro_historical test to two small unit tests

### LINKED ISSUE #115: Add tests for test experiment setup
In the tests for repro tests (e.g. `tests/test_test_bit_reproducibility.py`), there's some checks for the test setup which can be separated out (as suggested in this [PR](https://github.com/ACCESS-NRI/model-config-tests/pull/110#discussion_r1968501517)). There probably should be some  unit tests for `exp_test_helper.py` checking the setup methods in a separate test file. Similarly for each model's setup - e.g. `set_model_runtime`.

Qian-HuiChen:
Previously `test_test_repro_historical` is a mixed test of `setup_exp` and historical repro test.

This PR adds a separate unit tests on `setup_exp` only. This PR additionally adds a separate file to test `set_model_runtime` of ACCESS-ESM1.5, ACCESS-OM2, and ACCESS-OM3.

Closes #115.

### DIFF
--- a/tests/config_tests/test_model_set_model_runtime.py
+++ b/tests/config_tests/test_model_set_model_runtime.py
@@ -1,0 +1,160 @@
+import shutil
+from unittest.mock import Mock
+
+import f90nml
+import pytest
+import yaml
+from payu.models.cesm_cmeps import Runconfig
+
+from model_config_tests.models import index as model_index
+from tests.common import RESOURCES_DIR
+
+
+@pytest.fixture
+def model_factory(isolated_config, tmp_path):
+    """Factory fixture to create model instances with an isolated configuration."""
+
+    def factory(model_name, config_name, output):
+        resources_dir = RESOURCES_DIR / model_name
+
+        # Mock ExpTestHelper
+        mock_experiment = Mock()
+        mock_experiment.output000 = resources_dir / output
+        mock_experiment.restart000 = resources_dir / "restart000"
+
+        mock_experiment.control_path = tmp_path / "control"
+        _, config_dir = isolated_config(config_name)
+        shutil.move(config_dir, mock_experiment.control_path)
+
+        mock_experiment.config_path = tmp_path / "control" / "config.yaml"
+        with open(mock_experiment.config_path) as f:
+            mock_experiment.config = yaml.safe_load(f)
+
+        # Create Model instance
+        ModelType = model_index[model_name]
+        return ModelType(mock_experiment)
+
+    return factory
+
+
+@pytest.mark.parametrize(
+    "expected_runtime",
+    [
+        {"years": 1, "months": 2, "days": 3, "seconds": 0},
+        {"years": 0, "months": 1, "days": 5, "seconds": 0},
+    ],
+)
+def test_set_model_runtime_esm1p5(model_factory, expected_runtime):
+    """Test that ACCESS-ESM1p5 set_model_runtime sets the correct runtime in config.yaml"""
+    # Set up ACCESS-ESM1.5 model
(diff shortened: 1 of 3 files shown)


tests/config_tests/test_model_set_model_runtime.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+    # Check that the config.yaml file has been updated with the expected runtime
+    with open(model.experiment.config_path) as f:
+        updated_config = yaml.safe_load(f)
+    assert updated_config["calendar"]["runtime"] == expected_runtime
+
+    # Check that atmosphere has 48 timesteps per day
+    atmosphere_config = (
+        model.experiment.control_path / model.submodels["um"] / "namelists"
```
jo-basevi: With tests, I find it's normally good to hard code some values. For example if you hard code the directory names `model.submodels["um"]` -> `atmosphere` and `model.submodels["cice"]` -> `ice` so it's not using the same logic as the code we are testing and  it checks using the `model.submodels` dict also works as expected.

jo-basevi: Though feel free to leave as-is

tests/test_exp_test_helper.py
```diff
@@ -656,3 +657,44 @@ def test_setup_manifests_unchanged_show_changes(mock_run, exp):
...
+    [
+        # ACCESS-ESM1.5 models
+        ("access", "esm1p5-prein", "control", "control-test_exp"),
+        ("access", "esm1p5-prein", "base-experiment", "test_exp"),
+        # ACCESS-OM2 models
+        ("access-om2", "om2-1deg", "control", "control-test_exp"),
+        ("access-om2", "om2-1deg", "base-experiment", "test_exp"),
+        # ACCESS-OM3 models
+        ("access-om3", "om3-100km", "control", "control-test_exp"),
+        ("access-om3", "om3-100km", "base-experiment", "test_exp"),
```
jo-basevi: Probably only need a couple configurations to test this one as it doesn't have specific model configuration logic

jo-basevi (dismissed):
Thanks for adding more tests @Qian-HuiChen! Could remove the tests checking runtime and experiment configuration set up from `test_test_bit_reproducibility.py`? If so, would also need a test for checking default runtimes. 

Eventually could move tests into model specific files such as `models/test_access_om2.py` but probably fine for now to group similar tests in one file.

tests/config_tests/test_model_set_model_runtime.py
Qian-HuiChen: Sounds good. Changed and added comment in [bcfa59c](https://github.com/ACCESS-NRI/model-config-tests/pull/233/commits/bcfa59c80f54f6b434c417f37c4738a247cf1069)

Qian-HuiChen:
In [bcfa59c](https://github.com/ACCESS-NRI/model-config-tests/pull/233/commits/bcfa59c80f54f6b434c417f37c4738a247cf1069), I removed the runtime check and expt set up check from `test_test_bit_reproducibility.py`. I added the default runtime check in `test_model_set_model_runtime.py`.

tests/test_exp_test_helper.py
Qian-HuiChen: Oh good point. I simplify this test in commit [9fb49fb](https://github.com/ACCESS-NRI/model-config-tests/pull/233/commits/9fb49fba1343ac5219ee097e20989fd5480f461b)

jo-basevi (approved):
Looks good to me!

